PR #1, opened 2021-03-17
add data layer

dvd-schwrtz:


packages/data/queries/exchanges.ts
```diff
@@ -0,0 +1,17 @@
+import { gql } from 'graphql-request'
+
+export const exchangesQuery = gql`
+    query shorts($account: String!) {
```
dvd-schwrtz, line 4: just a placeholder while i set things up

JChiaramonte7 (commented):
Rate update queries look solid. Thanks!

PR #3, opened 2021-03-23
feat: NetworkID enums for ovm mainnet and kovan

JChiaramonte7:
Needed types for OVM chain IDs on kwenta

### DIFF
--- a/packages/contracts-interface/src/types.ts
+++ b/packages/contracts-interface/src/types.ts
@@ -17,6 +17,8 @@ export enum NetworkId {
 	Rinkeby = 4,
 	Goerli = 5,
 	Kovan = 42,
+	'Mainnet-Ovm' = 10,
+	'Kovan-Ovm' = 69,
 }
 
 type ContractInfo = {


packages/contracts-interface/src/types.ts
```diff
@@ -17,6 +17,8 @@ export enum NetworkId {
 	Rinkeby = 4,
 	Goerli = 5,
 	Kovan = 42,
+	'Mainnet-Ovm' = 10,
+	'Kovan-Ovm' = 69,
```
evgenyboxer, line 21: Just wondering if it's going to be a bit annoying to type it with quotes 👯 

Is `MainnetOvm` or `MainnetOVM` bad?

I guess I'm fine with either, just wondering if its easier to type while still understanding its OVM

evgenyboxer (approved):

packages/contracts-interface/src/types.ts
JChiaramonte7: Yeah I just ended up going with what we have for `Network` types a few lines up for consistency.

I think /synthetix repo also uses 'Kovan-Ovm' etc.

clementbalestrat: Correct. This is why we have to use the quotes.

PR #4, opened 2021-03-26
feat: parsing for data layer

JChiaramonte7:
Adds parsing for:
- synthetix
- issued
- synthExchanges


packages/data/queries/synthExchanges/parse.ts
```diff
@@ -0,0 +1,37 @@
...
+    toCurrencyKey,
+}: SynthExchange): SynthExchange => ({
+	account,
+    block: Number(block),
+    feesInUSD: ethers.utils.formatEther(feesInUSD),
+    from,
+    fromAmount: ethers.utils.formatEther(fromAmount),
+    fromAmountInUSD: ethers.utils.formatEther(fromAmountInUSD),
+    fromCurrencyKey: fromCurrencyKey,
+    gasPrice,
```
JChiaramonte7, line 29: left this in wei big number, not sure if wanted something else, but make sense to leave in wei if we're going to pass this on into a tx later or something

dvd-schwrtz (approved):

PR #21, opened 2021-05-10
update snx data to new versions in testing

KillerByte:
update tests and parser code for new proposed charts.

need to update repo to point back to `synthetix-team` instead of `killerbyte`

currently one failing test due to reduced rate update frequency samples, check needed on the subgraphs.


clementbalestrat:
@dbeal-eth this PR looks old. Should I close it?

PR #23, opened 2021-05-16
add new queries library

KillerByte:
refactors queries shared between staking and kwenta into a single lib, with some changes

tests have determined not to be added in this PR since they were not available in the original repos. we should decide whether we want to mock the data in these functions, or run E2E to establish broad functionality.

proposed notes/changes for implementation:
* targeting moving of these applications without significant modifications
* all queries are now expected to accept `QueryContext` as first argument, which includes
common initialized libs such as snxjs, ethers provider, snxData. this is intended to produce
effective network switching
* all amounts should now be expressed as `Wei`, which can be easily converted into using the
`wei()` helper function, and converted as necessary for display or math


packages/queries/package.json
```diff
@@ -0,0 +1,76 @@
...
+	},
+	"author": "Synthetix",
+	"license": "MIT",
+	"bugs": {
+		"url": "https://github.com/Synthetixio/js-monorepo/issues"
+	},
+	"dependencies": {
+		"@synthetixio/data": "../data",
+		"@synthetixio/wei": "../wei",
+		"axios": "^0.21.1",
```
dvd-schwrtz, line 35: let's lock all the packages and remove the `^`

packages/queries/src/currency.ts
```diff
@@ -0,0 +1,146 @@
+import keyBy from 'lodash/keyBy';
+import { Rates } from './queries/rates/useExchangeRatesQuery';
+
+export type CurrencyKey = string;
+
+export enum Synths {
```
dvd-schwrtz, line 6: hmmm @clementbalestrat what is the best way to manage this enum moving forward? We also have the assets repo too. Ideally it would be easy to manage lists like this without updating multiple places and if possible read something on chain and just remove the list? that way we don't have to update anything ever?

packages/queries/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+type RawSynthetixQueries = typeof FUNCS;
+
+type OmitFirstArg<F> = F extends (x: any, ...args: infer P) => infer R ? (...args: P) => R : never;
+
+export type SynthetixQueries = {
+    [Property in keyof RawSynthetixQueries]: OmitFirstArg<RawSynthetixQueries[Property]>;
+};
+
+export default async function synthetixQueries({ provider }: { provider: ethers.providers.Provider }): Promise<SynthetixQueries> {
```
dvd-schwrtz, line 75: I like this a lot but it is probably pretty confusing for those not familiar with typescript. We should just add some comments explaining how this file works and should be good

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -0,0 +1,74 @@
...
+	average: number;
+};
+
+export type GasSpeed = keyof GasPrices;
+
+export const GAS_SPEEDS: GasSpeed[] = ['average', 'fast', 'fastest'];
+
+const useEthGasPriceQuery = (ctx: QueryContext, options?: UseQueryOptions<GasPrices>) => {
+	return useQuery<GasPrices>(
+		['network', 'gasPrice', ctx.network],
```
dvd-schwrtz, line 46: nice

clementbalestrat (commented):
Great stuff @KillerByte

packages/queries/src/currency.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export const ETH_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
+
+export const sUSD_EXCHANGE_RATE = 1;
+
+export const CRYPTO_CURRENCY = [
+	'KNC',
+	'COMP',
+	'REN',
+	'LEND',
```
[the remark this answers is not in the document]
dvd-schwrtz: not sure. let's double check before we merge this

packages/queries/src/queries/rates/useSynthMarketCapQuery.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const useSynthMarketCapQuery = (
+	ctx: QueryContext,
+	currencyKey: CurrencyKey | null,
+	options?: UseQueryOptions<Wei>
+) => {
+	return useQuery<Wei>(
+		['rates', 'marketCap', ctx.networkId, currencyKey],
+		async () => {
+			return wei(await ctx.snxjs!.contracts[synthToContractName(currencyKey!)].totalSupply());
```
dvd-schwrtz, line 16: total supply is not the same as the market cap

packages/queries/src/currency.ts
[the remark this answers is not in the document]
dvd-schwrtz: yes, let's add 'AAVE' we might have to look through and add other synths as well. Let's do some changes around how the synths are managed in another PR

packages/queries/src/queries/rates/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+	const oldPrice = rates[rates.length - 1].rate;
+	const percentageChange = (newPrice - oldPrice) / oldPrice;
+
+	return percentageChange;
+};
+
+export const calculateTimestampForPeriod = (periodInHours: number) =>
+	Math.trunc(subHours(new Date().getTime(), periodInHours).getTime() / 1000);
+
+export const mockHistoricalRates = (
```
dvd-schwrtz, line 34: maybe let's change the name of this function to `usdHistoricalRates` instead of mock

packages/queries/src/currency.ts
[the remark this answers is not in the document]
KillerByte: will add `AAVE` and any other synths that I notice are missing :+1:

packages/queries/src/queries/rates/useSynthMarketCapQuery.ts
KillerByte: copied this directly from kwenta https://github.com/Synthetixio/kwenta/blob/dev/queries/rates/useSynthMarketCapQuery.ts#L21

which is better, should the `use` function be renamed, or should the functionality be corrected?

packages/queries/src/currency.ts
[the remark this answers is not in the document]
KillerByte: i assume so. the list is from here: https://github.com/Synthetixio/staking/blob/dev/constants/currency.ts#L3

this stuff might be best generated from snxjs somehow? thats effectively what needs to be synced up, from my brief analysis.

packages/queries/src/index.ts
KillerByte: comments added

packages/queries/src/currency.ts
KillerByte: I mentioned in a separate note, but probably best is to use the source of truth, which is probably 

the challenge might be dealing with the fact its an `enum`. If we can convert it to a `type`, or that is the only type of usage required, we can import the json file from synthetix https://github.com/Synthetixio/synthetix/blob/develop/publish/deployed/mainnet/synths.json and use type computation to establish the same effect.

clementbalestrat: Could we just build this file during the build process?
- remove this file
- fetch latest synths.json from Synthetix
- rebuild the file with latest changes

clementbalestrat: To be fair I'd be tempted to put this file into a new lib called `currency`  or `constants` so we can use it in the dApps. I think we're hardcoding this list pretty much everywhere right @dvd-schwrtz ?

packages/queries/src/queries/rates/useSynthMarketCapQuery.ts
clementbalestrat: We should probably rename it to `useSynthTotalSupplyQuery` then. Or fetch the asset price and multiply it by the total supply to have the market cap. It depends on how we are using it in the dApps.

I believe we do need the total supply sometimes, so maybe let's just rename it.

clementbalestrat (approved):

PR #28, opened 2021-06-18
fix: add missing negative entries in debt pool share compute

vbstreetz:


KillerByte (changes requested):
we talked in private conversation and agreed since rest of queries uses `wei` library, numbers exported by this function should also use `wei`.

vbstreetz:
> we talked in private conversation and agreed since rest of queries uses `wei` library, numbers exported by this function should also use `wei`.

awesome. on it..

vbstreetz:
> > we talked in private conversation and agreed since rest of queries uses `wei` library, numbers exported by this function should also use `wei`.
> 
> awesome. on it..

hey! i just pushed an update to fix this.

vbstreetz:
This has been successfully integrated in stats https://stats-git-use-synthetix-queries-pkg-synthetixio.vercel.app

Screenshot 2021-07-30 at 18 24 42

JChiaramonte7 (approved):
lgtm

packages/queries/__tests__/synths.test.ts
```diff
@@ -11,6 +12,10 @@ describe('@synthetixio/queries synths', () => {
 	test('useSynthsTotalSupplyQuery', async () => {
 		const wrapper = getWrapper();
 
+		const [sETHKey, sBTCKey, sUSDKey] = [Synths.sETH, Synths.sBTC, Synths.sUSD].map(
```
KillerByte, line 15: this constant isnt great, but its probably fine for now.

KillerByte: instead it should let the end user application send in the list of keys they want, not hardcoded here.

packages/queries/package.json
```diff
@@ -34,7 +34,7 @@
 	"dependencies": {
 		"@snapshot-labs/snapshot.js": "^0.1.11",
 		"@synthetixio/contracts-interface": "^2.47.0-ovm.4",
-		"@synthetixio/data": "^2.47.0-ovm.4",
+		"@synthetixio/data": "0.0.0-3bc863c",
```
KillerByte, line 37: make sure this doesn't  get pushed

KillerByte (approved):
please fix a couple things but overall really appreciate :+1:

packages/queries/package.json
vbstreetz: plan is to resolve this in a follow-up PR..

PR #30, opened 2021-06-22
Add local networks

liamzebedee:
Epic: https://github.com/Synthetixio/synthetix/pull/1356

### DIFF
--- a/packages/optimism-networks/src/constants.ts
+++ b/packages/optimism-networks/src/constants.ts
@@ -3,11 +3,13 @@ import { OptimismNetwork, NetworkMapper, LayerTwoNetworkId, MessengerAddress } f
 export const L1_TO_L2_NETWORK_MAPPER: NetworkMapper = {
 	1: 10,
 	42: 69,
+	31337: 420,
 };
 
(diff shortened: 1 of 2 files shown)


packages/optimism-networks/src/constants.ts
```diff
@@ -31,6 +33,16 @@ export const OPTIMISM_NETWORKS: Record<number, OptimismNetwork> = {
 			'https://optimism.io/images/metamask_icon.png',
 		],
 	},
+	420: {
```
liamzebedee, line 36: TODO: had a chat with Clem. We will probably put this in a guide - apparently Metamask's API complains if the rpc url is http, which it will be for the local optimism ops node.

JChiaramonte7 (approved):
glorious

liamzebedee:
@JChiaramonte7 glory be

looks like there's a timeout error [here](https://github.com/Synthetixio/js-monorepo/pull/30/checks?check_run_id=3823242267#step:8:159):

> ● @synthetixio/data tests › debtSnapshots query › should return debtSnapshots data from l1
> thrown: "Exceeded timeout of 5000 ms for a test.

Is this related or extraneous (eg. intermittent)?

PR #42, opened 2021-07-19
fix(data): added codegen step to data package build

JChiaramonte7:
Just makes sure that we generate our types before building


JChiaramonte7:
covered by #46

PR #43, opened 2021-07-19
modify context to take a signer and to update the synthetix-js and sy…

dvd-schwrtz:
…nthetix-data instances when the network changes


packages/queries/src/context.ts
```diff
@@ -8,4 +8,13 @@ export interface QueryContext {
 	provider: ethers.providers.Provider | null;
 	snxData: SynthetixData | null;
 	snxjs: SynthetixJS | null;
+	updateQueryContext?: ({
```
KillerByte, line 11: i don't feel great about this mostly because it circumvents the react state tracking functions. make sure we are knowing what we do here before we actually do it is all.

packages/queries/src/context.ts
```diff
@@ -8,4 +8,13 @@ export interface QueryContext {
 	provider: ethers.providers.Provider | null;
```
KillerByte, line 8: perhaps `QueryContext` should be converted to a fully fledged class if we are going to start adding state update functions to it?

KillerByte (approved):
just be careful

packages/queries/src/QueryWrapper.tsx
```diff
@@ -0,0 +1,97 @@
...
+type SynthetixQueryWrapperState = {
+	queryContext: QueryContext;
+};
+
+class SynthetixQueryWrapper extends Component<
+	SynthetixQueryWrapperProps,
+	SynthetixQueryWrapperState
+> {
+	state: SynthetixQueryWrapperState = {
+		// optional second annotation for better type inference
```
dvd-schwrtz, line 58: will delete this comment

packages/queries/src/index.ts
```diff
@@ -59,45 +57,18 @@ export type SynthetixQueries = {
...
-	}
-
-	return ctx;
-}
-
-const SynthetixQueryContext = React.createContext<QueryContext | null>(null);
-export const SynthetixQueryContextProvider = SynthetixQueryContext.Provider;
+export { SynthetixQueryWrapper, SynthetixQueryContext };
 
 export default function useSynthetixQueries(): SynthetixQueries {
```
KillerByte, line 62: it has been discussed to consider moving this function to within the `SynthetixQueryWrapper` and remapping this function to just read from it to get hte list of functions (instead of re-mapping them every time)

PR #44, opened 2021-07-20
feat: added l2 gas query

JChiaramonte7:
Extended our useEthGasPriceQuery to incorporate ovm gas querying

### DIFF
--- a/packages/queries/src/queries/network/useEthGasPriceQuery.ts
+++ b/packages/queries/src/queries/network/useEthGasPriceQuery.ts
@@ -32,10 +32,22 @@ type GasNowResponse = {
 	};
 };
 
-const useEthGasPriceQuery = (ctx: QueryContext, options?: UseQueryOptions<GasPrices>) => {
+const useEthGasPriceQuery = (
+	ctx: QueryContext,
+	useOVM?: boolean,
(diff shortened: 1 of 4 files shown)


packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -32,10 +32,22 @@ type GasNowResponse = {
...
+const useEthGasPriceQuery = (
+	ctx: QueryContext,
+	useOVM?: boolean,
+	options?: UseQueryOptions<GasPrices>
+) => {
 	return useQuery<GasPrices>(
 		['network', 'gasPrice', ctx.networkId],
 		async () => {
+			if (useOVM) {
+				const gasPrice = (await ctx.provider!.getGasPrice()).toNumber() / 1e8 / 10 ?? 0;
```
clementbalestrat, line 44: since we're doing an async call it's probably better to move this part into a try catch block

clementbalestrat: also `1e8/10` should probably be a function sitting in a utils file. Like formatGwei or something

clementbalestrat (changes requested):

clementbalestrat (approved):

PR #45, opened 2021-07-22
add queries from staking app

KillerByte:
all changes:
* add queries from staking app
* have queries use codegen to link query function dependencies
* fix internal dependencies by lerna to use exact version (otherwise it uses the highest hash)
* fix modules to always export `module` definition
* add dependencies for `@synthetixio/data` for new queries


packages/queries/jest.config.js
```diff
@@ -0,0 +1,13 @@
+/** @type {import('@ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+	preset: 'ts-jest',
+	testEnvironment: 'node',
+	coverageThreshold: {
+		global: {
+			branches: 0,
```
dvd-schwrtz, line 7: do we need the config if thresholds are 0?

package.json
```diff
@@ -50,5 +51,10 @@
 		"hooks": {
 			"pre-commit": "npm run lint"
 		}
+	},
```
clementbalestrat, line 54: what's an optional dependency?

packages/data/queries/rateUpdates/parse.ts
```diff
@@ -8,7 +7,8 @@ export const parseRates = (rate: RateUpdate, networkId?: number): RateUpdate =>
 	const parsedRate =
 		networkId && networkId === NetworkId['Kovan-Ovm']
 			? rateValue
-			: ethers.utils.formatEther(rateValue);
+			: // todo: I have no idea what this code is supposed to do
+			  (rateValue / 1e18).toString();
```
clementbalestrat, line 11: let's just get rid of it

packages/queries/src/constants.ts
```diff
@@ -71,3 +71,5 @@ export const PERIOD_LABELS_MAP: Record<Period, PeriodLabel> = {
 };
 
 export const PERIOD_LABELS = Object.values(PERIOD_LABELS_MAP);
+
+export const COUNCIL_NOMINATIONS_URL = 'https://';
```
clementbalestrat, line 75: whats this URL for exactly?

clementbalestrat (approved):
LGTM

package.json
KillerByte: npm now complains if its not included. maybe newer version or somethign? I think its a dependency that, if someone were to include the repository root (which they wouldn't), they can "optionally" install the dependency.

it did have a positive effect, not just on the pipeline, but also on my computer running pre-commit hooks, so its worthwhile to include them.

packages/data/queries/rateUpdates/parse.ts
KillerByte: can't becuase the test will fail otherwise

KillerByte: we are probably going to replace the `@synthetixio/data` repo with a code generator in the near future so I'm not super keen on carefully maintaining this for the time being.

packages/queries/src/constants.ts
KillerByte: its depended upon by one of the queries. Let me check what the actual value for that is supposed to be and fill it in.

PR #47, opened 2021-07-26
feat(data): Add L2 kovan rates to rateUpdates

kmeraz:
- Adding the endpoint for L2 Kovan rate updates
- This will allow Kwenta to render the rates chart for rates on L2 Kovan

### DIFF
--- a/packages/data/queries/rateUpdates/parse.ts
+++ b/packages/data/queries/rateUpdates/parse.ts
@@ -1,19 +1,21 @@
 import { ethers } from 'ethers';
 import { RateUpdate } from '../../generated/graphql';
 import { formatTimestamp } from '../../src/utils';
+import { NetworkId } from '@synthetixio/contracts-interface';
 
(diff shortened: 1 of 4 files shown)


packages/data/queries/rateUpdates/parse.ts
```diff
@@ -1,19 +1,21 @@
...
-	block: Number(block),
-	currencyKey,
-	id,
-	rate: ethers.utils.formatEther(rate),
-	synth,
-	timestamp: formatTimestamp(timestamp),
-});
+export const parseRates = (rate: RateUpdate, networkId?: number): RateUpdate => {
+	const { block, currencyKey, id, rate: rateValue, synth, timestamp } = rate;
+	const parsedRate =
```
JChiaramonte7, line 8: @KillerByte temporary conditional we can remove after the L1 subgraphs are updated

JChiaramonte7 (approved):

PR #50, opened 2021-07-28
WIP: Balances

dvd-schwrtz:


dbeal-eth:
this PR is boing to be closed since balances implementation is a bit different now, and we have the `codegen-graphql-ts` replacing this library entirely.

PR #53, opened 2021-08-02
fix: unspecify block height from batched calls

JChiaramonte7:


kmeraz (approved):

JChiaramonte7:
TODO: mock out ethcall and add another PR with tests for the token balances query

PR #56, opened 2021-08-05
add data README

dbeal-eth:
just a really simple readme for reference, but later on this library should be deprecated so didn't want to get too detailed.


dbeal-eth:
not needed since we will be using the generator library soon

PR #59, opened 2021-08-16
fix(queries): added decimal argument to wei conversion in token balance query

JChiaramonte7:
We were converting all currency balances to wei with the default 18 decimal standard. This fixes the conversation for tokens that don't use 18 decimals (ie. USDC, USDT)

### DIFF
--- a/packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
+++ b/packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
@@ -50,7 +50,7 @@ const useTokensBalancesQuery = (
 				const token = tokensMap[symbol];
 
 				return {
-					balance: wei(balance),
+					balance: wei(balance, token.decimals, true),
 					token,
 				};
 			});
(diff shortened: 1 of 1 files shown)


kmeraz (approved):

packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
```diff
@@ -50,7 +50,7 @@ const useTokensBalancesQuery = (
 				const token = tokensMap[symbol];
 
 				return {
-					balance: wei(balance),
+					balance: wei(balance, token.decimals, true),
```
KillerByte, line 53: is `token.decimals` nullable? if so, should add default to 18. in fact, might want to declare the input type here at this point rather than just leaving it as `any`.

also, `wei` automatically interprets `BigNumber` types to be `asWei = true` already, so specifying the last argument is not necessary. typically I only provide the argument as an override

KillerByte (changes requested):

packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
JChiaramonte7: fixed and going to merge so i can fix kwenta, but if fixes weren't sufficient let me know

PR #80, opened 2021-09-17
[optimism-networks] fix: remove 'fraudProofWindow' from optimism network config

kmeraz:
Metamask throws an error when trying to use optimism via the current config.

image


JChiaramonte7 (approved):
re-running CI

kmeraz:
@JChiaramonte7 i just took this for a spin on kwenta and it's lookin good

PR #85, opened 2021-09-23
fix(useEVMTxn): disable gas estimate if tx disabled

vbstreetz:


packages/queries/src/mutations/useContractTxn.ts
```diff
@@ -11,7 +11,7 @@ const useContractTxn = (
 	txnOptions: Partial<ethers.providers.TransactionRequest> = {},
 	options?: UseEVMTxnOptions
 ) => {
-	if (contract != null && method != null) {
+	if (contract != null && method != null && options?.enabled) {
```
dbeal-eth, line 14: why is this needed? also what happensi f `options` is not defined (as it often isnt)?

packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -60,6 +60,10 @@ const useEVMTxn = (
 
 		setErrorMessage(null);
 
+		if (!options.enabled) {
```
dbeal-eth, line 63: would probably be better to just do `if(options.enabled) estimateGas().then()...`

dbeal-eth (changes requested):

packages/queries/src/mutations/useContractTxn.ts
vbstreetz: this breaks if args are not valid https://github.com/Synthetixio/js-monorepo/blob/master/packages/queries/src/mutations/useContractTxn.ts#L19

packages/queries/src/mutations/useEVMTxn.ts
vbstreetz: i like early returns but fixed :)

packages/queries/src/mutations/useContractTxn.ts
dbeal-eth: really? interesting. good catch :+1:

dbeal-eth (approved):
there are a couple "while you are at it" changes I would like to request but one thing at a time is probably best.

thanks for your work on this :+1:

vbstreetz:
> there are a couple "while you are at it" changes I would like to request but one thing at a time is probably best.
> 
> thanks for your work on this 👍

say more?

packages/queries/src/mutations/useContractTxn.ts
vbstreetz: btw, i updated to https://github.com/Synthetixio/js-monorepo/blob/fix/useEvmTxn/packages/queries/src/mutations/useContractTxn.ts#L14

dbeal-eth:
if you have time/desire, some fixes that would be helpful are:
* fixing `UseEVMOptions` such that all the arguments are optional instead of required all at once
* improving errored transaction handling (right now its not always good at parsing the errors to set for `txn.errorMessage`)
* adding EIP-1559 support (which would add EIP-1559 to all of our downstream apps, which would be nice)

packages/queries/src/mutations/useContractTxn.ts
vbstreetz: > if you have time/desire, some fixes that would be helpful are:
> 
> * fixing `UseEVMOptions` such that all the arguments are optional instead of required all at once
> * improving errored transaction handling (right now its not always good at parsing the errors to set for `txn.errorMessage`)
> * adding EIP-1559 support (which would add EIP-1559 to all of our downstream apps, which would be nice)

awesome. tackling these in a new or..

PR #105, opened 2021-10-15
Fix/gas estimate

0xclem:


bachstatter (approved):
LGMT. Not sure how the monorepo stuff works yet, maybe you need to increase version somewhere?

PR #109, opened 2021-10-18
Avoid infinity loop

bachstatter:
I think we only need to update the `gasLimit` when the `to` address changes?

With the previous implementation we were in an infinity loop. Fetching gas meant that the transaction got its `data` and `value` set. When it finished those properties gets set to null causing the useEffect to be triggered again.

### DIFF
--- a/packages/queries/src/mutations/useEVMTxn.test.ts
+++ b/packages/queries/src/mutations/useEVMTxn.test.ts
@@ -5,7 +5,7 @@ import { getWrapper } from '../../testUtils';
 import useEVMTxn from './useEVMTxn';
 
 describe('useEVMTxn', () => {
-	test('Does not call estimateGas when tnx is null', async () => {
+	test('Does not call estimateGas when txn is null', async () => {
 		const estimateGasMock = jest.fn().mockResolvedValue(10);
 		const sendTransactionMock = jest.fn();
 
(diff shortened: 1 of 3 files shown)


clementbalestrat:
@bachstatter but if you actually change the value or the data, then it won't refresh?

Example: You want to burn 10 sUSD. It gets the gasLimit for you. Then you type another 0 in the input to make it 100 sUSD. I guess it won't refresh it?

cc @dbeal-eth

packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -74,7 +74,7 @@ const useEVMTxn = (
 		}
 	}
 
-	useEffect(refresh, [txn?.data, txn?.value, txn?.nonce, txn?.from, txn?.to]);
```
dbeal-eth, line 77: choosing to refresh on all of these properties was a deliberate choice. if any of these values are changed, it could affect the outcome/execution of the transaction, and therefore we need to calculate an updated gasLimit/other derivative data.

bachstatter:
Thanks for the tips @dbeal-eth, I though we got the infinity loop by the transaction changing `data` from having a value and being `null`. But I saw now that we have a guard for null. So will digg a bit more, not super easy since the logs are getting spammed 😅

packages/queries/src/mutations/useEVMTxn.test.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { renderHook } from '@testing-library/react-hooks';
+import { BigNumber } from 'ethers';
+import { getWrapper } from '../../testUtils';
+import useEVMTxn from './useEVMTxn';
+
+describe('useEVMTxn', () => {
+	test('Does not call estimateGas when tnx is null', async () => {
```
dbeal-eth, line 8: np "Does not call estimateGas when txn is null"

PR #110, opened 2021-10-18
Fix/use token balances query

0xjocke:
-  The passed in tokens needs to be part of the cache key.  One part of the consumer app might pass just eth and in another context another token. Without having tokens as a key it would still only give back result for eth
- Improve and declare return type
- Remove as, tsignore, lodash helpers and simplify mapping

### DIFF
--- a/packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
+++ b/packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
@@ -1,54 +1,42 @@
-import { useQuery, UseQueryOptions } from 'react-query';
+import { useQuery, UseQueryOptions, UseQueryResult } from 'react-query';
 import { wei } from '@synthetixio/wei';
 import erc20Abi from '../../abis/ERC20.json';
-import zipObject from 'lodash/zipObject';
-import omitBy from 'lodash/omitBy';
-import mapValues from 'lodash/mapValues';
-import keyBy from 'lodash/keyBy';
-import { ethers } from 'ethers';
-
+import { BigNumber, ethers } from 'ethers';
 import { CRYPTO_CURRENCY_MAP } from '../../currency';
-
 import { QueryContext } from '../../context';
 import { Token, TokenBalances } from '../../types';
 
+type UseTokensBalancesQueryReturn = UseQueryResult<TokenBalances>;
 const useTokensBalancesQuery = (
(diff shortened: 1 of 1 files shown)


packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
```diff
@@ -1,54 +1,42 @@
...
-			return mapValues(positiveBalances, (balance: ethers.BigNumber, symbol: string) => {
-				const token = tokensMap[symbol];
-
-				return {
-					balance: wei(balance, token.decimals ?? 18),
-					token,
-				};
+			};
+			const promises = tokens.map(async (token) => {
+				if (!ctx.provider || !walletAddress) return { balance: wei(0), token };
```
dbeal-eth, line 29: this is unneeded since the query is only enabled iff `!!ctx.provider && tokens.length > 0 && !!walletAddress`

packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
```diff
@@ -1,54 +1,42 @@
...
+				const balance = await getBalance(token);
+				return { balance: wei(balance, token.decimals ?? 18), token };
 			});
+
+			const data = await Promise.all(promises);
+			return data.reduce((acc: TokenBalances, val) => {
+				if (val.balance.lte(0)) return acc;
```
dbeal-eth, line 36: this seems lnike a pretty significant change in behavior, not returning 0 balances seems rather unnecessary since the app can do that itself, but it would be harder to re-add the 0 balances if they weren't returned

packages/queries/src/queries/walletBalances/useTokensBalancesQuery.ts
```diff
@@ -1,54 +1,42 @@
...
 const useTokensBalancesQuery = (
 	ctx: QueryContext,
 	tokens: Token[],
 	walletAddress: string | null,
 	options?: UseQueryOptions<TokenBalances>
-) => {
-	const symbols = tokens.map((token) => token.symbol);
-	const tokensMap = keyBy(tokens, 'symbol');
-
+): UseTokensBalancesQueryReturn => {
```
dbeal-eth, line 15: np but I believe this value should be calculable by TypeScript type inference? were you not getting any type chceking when using this on an end app?

Issue #115, opened 2021-11-01
codegen-graph-ts not working with next apps


bachstatter:
When using the new version (`2.52.1-ovm`) of Synthetix Queries with `codegen-graph-ts`  in a Next app I get this error:
`ReferenceError: global is not defined`

Is `codegen-graph-ts` suppose to be working from the frontend? 
It depends on `node-fetch` which contains the line that is failing:
`Body.Promise = global.Promise;`
![Screen Shot 2021-11-01 at 4 49 57 pm](https://user-images.githubusercontent.com/5688912/139628007-0964008e-bafd-4a18-be07-336587847bbf.png)

PR #116, opened 2021-11-01
update queries with latest version of graph codegen

dbeal-eth:
this release uses axios, which fixes queries from within the browser


bachstatter:
I think this closes #115

Issue #117, opened 2021-11-02
update useEthGasPriceQuery to support EIP1559 and deprecate GASNOW API


fritzschoff:
The hook `useEthGasPriceQuery` is using the `gasnow` api which is not active anymore.
Also @dbeal-eth pointed out that this hook needs to be refactored due to EIP1559.

PR #118, opened 2021-11-02
remove 'gas now' API from gas price query

kmeraz:


fritzschoff:
ref #117

fritzschoff (approved):

packages/queries/__tests__/network.test.ts
```diff
@@ -53,4 +59,50 @@ describe('@synthetixio/queries network useEthGasPriceQuery', () => {
...
+
+		await waitFor(() => result.current.isSuccess);
+
+		expect(axios.get).toBeCalledWith(ETH_GAS_STATION_API_URL);
+	});
+
+	test('should use the provided provider for not Mainnet network ', async () => {
+		const newCTX = getFakeQueryContext(NetworkId['Mainnet-Ovm']);
+		set(
+			newCTX.provider!,
```
0xclem, line 87: @kmeraz can you please fix the linter warnings?

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -3,11 +3,13 @@ import { useQuery, UseQueryOptions } from 'react-query';
 import { QueryContext } from '../../context';
 import { GasPrices } from '../../types';
 import { formatGwei } from '../../utils';
+import { BigNumber } from 'ethers';
```
0xclem, line 6: I don't think we should be using this lib since we've got the `Wei` one now

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -24,44 +26,45 @@ type EthGasStationResponse = {
...
 
+			try {
+				const block = await ctx?.provider?.getBlock('latest');
+				// if block has baseFeePerGas then networks supports EIP1559
+				if (block?.baseFeePerGas) {
 					return {
-						fastest: Math.round(fastest / 10),
-						fast: Math.round(fast / 10),
-						average: Math.round(average / 10),
+						fastest: block.baseFeePerGas
```
0xclem, line 53: It's great that you added a EIP 1159 support. However what is this formula? You are multiplying `baseFeePerGas` by 2 and then adding some wei on top of that?

0xclem (changes requested):

packages/queries/src/queries/network/useEthGasPriceQuery.ts
fritzschoff: The formula is from a blocknative blog post. 2 times blockBasFee with some premium for the miner. I didn't want to create two return types, that is why I squeezed that into this type structure. But that is highly questionable

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -5,9 +5,10 @@ import { GasPrices } from '../../types';
 import { formatGwei } from '../../utils';
 
 import { NetworkId } from '@synthetixio/contracts-interface';
+import Wei, { wei } from '@synthetixio/wei';
 
-const ETH_GAS_STATION_API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
-const GAS_NOW_API_URL = 'https://www.gasnow.org/api/v3/gas/price?utm_source=kwenta';
+export const ETH_GAS_STATION_API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
+const MULTIPLIER = new Wei(2, 1).toBN();
```
bachstatter, line 11: I think we usually just do use `wei()` rather than `new Wei()` def no biggie 😅

bachstatter (approved):

bachstatter:
Looks like a test is failing too, not sure if it's an actual mistake or assert error

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -24,44 +25,36 @@ type EthGasStationResponse = {
...
 
+			try {
+				const block = await ctx?.provider?.getBlock('latest');
+				// if block has baseFeePerGas then networks supports EIP1559
+				if (block?.baseFeePerGas) {
 					return {
-						fastest: Math.round(fastest / 10),
-						fast: Math.round(fast / 10),
-						average: Math.round(average / 10),
+						fastest: block.baseFeePerGas.mul(MULTIPLIER).add(wei(6, 9).toBN()).toNumber(),
```
0xclem, line 52: Ok so I had a look at the article from Blocknative which explains this method:  https://www.blocknative.com/blog/eip-1559-fees

I'm happy with that, even though the `priority fee` is a bit random (2, 4 or 6). But unfortunately there's no way to really predict it, except using blocknative API maybe?

https://www.blocknative.com/gas-estimator

Another thing. Before merging this PR, we need to change the structure of this object a bit because I believe (correct me if I'm wrong) there is no way for the frontend to now which one to use (EIP1559 or before).

The object always returns the same `{average: value, fast: value, fastest: value}` and this isn't correct.

For EIP1559 it should probably be something like:

`{average: { maxFeePerGas:  block.baseFeePerGas.mul(MULTIPLIER).add(wei(2, 9).toBN()).toNumber()     maxPriorityFeePerGas: wei(2, 9).toBN().toNumber()}, fast: ...etc, fastest: ...etc}`

Because the params for an EIP1559 transaction are different. And the frontend needs to know which ones to pass in a call.

0xclem (changes requested):

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -5,9 +5,10 @@ import { GasPrices } from '../../types';
 import { formatGwei } from '../../utils';
 
 import { NetworkId } from '@synthetixio/contracts-interface';
+import { wei } from '@synthetixio/wei';
 
-const ETH_GAS_STATION_API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
-const GAS_NOW_API_URL = 'https://www.gasnow.org/api/v3/gas/price?utm_source=kwenta';
+export const ETH_GAS_STATION_API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
+const MULTIPLIER = wei(2, 1).toBN();
```
dbeal-eth, line 11: note: this value would be `20`, not 2 as I imagine you would expect. So, if the gas is `150 gwei`, that would be multiplying by 20 to `3000 gwei` right?

dbeal-eth (changes requested):
I think there is a math error in the code here with `MULTIPLIER`. Also, I think we should be returning `maxBaseFee` and `maxPriorityFee` separately, since you don't really get 90% of the benefit of 1559 without having them separate. I also think we should do-away with `fast`, `fastest`, and `average` and just have. If the user want's more speed for some reason, almost every wallet supports their own system.

I'm sure you already know, but when having these values separately, during normal operation, the miner gets paid the `maxPriorityFee`, while the transaction sender only pays whatever the `baseFee` was at the block the transaction got included in. Therefore, the user saves `maxBaseFee - baseFee` gas. If you don't keep these values separate, the user ends up paying the full amount of `maxBaseFee` since the miner gets paid the full remainder.

PR #119, opened 2021-11-04
map subgraph functions with context same as rest of query library

dbeal-eth:
after implementing downstream, it has become apparent that resolving and passing the `subgraphUrl` every
time is not a very good pattern. As the subgraph data must be injected into the context for the queries library anyway, this will
allow for calling of subgraph functions directly from `useSynthetixQueries` with the first `subgraphUrl` argument
mapped with the url of the corresponding subgraph.

### DIFF
--- a/packages/queries/src/index.ts
+++ b/packages/queries/src/index.ts
@@ -31,7 +31,7 @@ type OmitFirstArg<F> = F extends (x: any, ...args: infer P) => infer R ? (...arg
 
 type Queries<T> = {
 	[Property in keyof T]: OmitFirstArg<T[Property]>;
-}
+};
 
 export type SynthetixQueries = {
 	exchanges: Queries<typeof exchanges>;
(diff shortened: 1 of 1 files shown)


bachstatter (commented):
Very nice, some impressive fiddling with `any`s back to proper types.

In [createQueryContext](https://github.com/Synthetixio/js-monorepo/blob/fix-codegen-args/packages/queries/src/index.ts#L51) do you think we can remove the `subgraphEndpoints` argument to synthetix constructor? 
We should be able to just create it by  `DEFAULT_SUBGRAPH_ENDPOINTS[networkId]` or  `DEFAULT_SUBGRAPH_ENDPOINTS[1]` when we're missing `networkId`

0xclem (approved):

0xclem:
@dbeal-eth LGTM, I agree with @bachstatter's comment

PR #122, opened 2021-11-10
Ovm gas fees

0xclem:


packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -35,19 +37,36 @@ const useEVMTxn = (
 	options?: UseEVMTxnOptions
 ) => {
 	const [gasLimit, setGasLimit] = useState<Wei | null>(null);
+	const [optimismLayerOneFee, setOptimismLayerOneFee] = useState<Wei | null>(null);
 	const [errorMessage, setErrorMessage] = useState<string | null>(null);
 	const [hash, setHash] = useState<string | null>(null);
 	const [txnStatus, setTxnStatus] = useState<TransactionStatus>('unsent');
 
-	async function estimateGas() {
+	const getOptimismLayerOneFees = async () => {
```
0xclem, line 45: I have moved the layerOneFee logic for Optimism in this file, which I then return so it can be used in the frontend

dbeal-eth (approved):

bachstatter:
CI was failing due to the version of `@eth-optimism/contracts` we used was missing types. I also think you did an npm install with an old version.
So, I reran `npm i` to get the package-lock back to v2 and upgraded `@eth-optimism/contracts` to latest version `0.4.14` -> `0.5.1`.

Types seems to be happy with the way we're using it but I didn't do any testing

fritzschoff (approved):

packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -34,52 +42,75 @@ const useEVMTxn = (
 	txn: ethers.providers.TransactionRequest | null,
 	options?: UseEVMTxnOptions
 ) => {
-	const [gasLimit, setGasLimit] = useState<Wei | null>(null);
+	const [gasLimit, setGasLimit] = useState<ethers.BigNumber | null>(null);
+	const [transactionFees, setTransactionFees] = useState<TransactionFees | Record<string, unknown>>(
```
bachstatter, line 46: This never get set, should it be removed?

packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -94,17 +125,15 @@ const useEVMTxn = (
 			const execTxn = clone(txn!);
 
 			try {
-				if (!execTxn.gasLimit && !OPTIMISM_NETWORKS[ctx.networkId!]) {
+				if (!execTxn.gasLimit) {
 					if (!gasLimit) {
 						const newGasLimit = (await estimateGas())!;
 						execTxn.gasLimit = newGasLimit?.mul(
 							1 + (options?.gasLimitBuffer || DEFAULT_GAS_BUFFER)
 						);
```
bachstatter, line 133: I think we need to do something smarter here. `newGasLimit` is an ethers bignumber. If we multiple that with 1.15 (ie. `DEFAULT_GAS_BUFFER`) we get decimals in some cases and that causes bignumber to throw 
`Uncaught Error: underflow (fault="underflow", operation="BigNumber.from", value=1.15, code=NUMERIC_FAULT, version=bignumber/5.4.2)`

bachstatter: Example when minting on mainnet:
Gas limit: `549284`
`549284 * 1.15 = 631676.6` -> bignumber throws

fritzschoff:
close #117

packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -34,52 +42,71 @@ const useEVMTxn = (
 	txn: ethers.providers.TransactionRequest | null,
 	options?: UseEVMTxnOptions
 ) => {
-	const [gasLimit, setGasLimit] = useState<Wei | null>(null);
+	const [gasLimit, setGasLimit] = useState<ethers.BigNumber | null>(null);
```
fritzschoff, line 45: why are we switching from `Wei` to ethers bignumber?

packages/queries/src/mutations/useEVMTxn.ts
```diff
@@ -34,52 +42,71 @@ const useEVMTxn = (
...
+		try {
+			const OptimismGasPriceOracleContract = new ethers.Contract(
+				optimismOracleContract.address,
+				optimismOracleContract.abi,
+				ctx.provider!
+			);
+			const serializedTxn = ethers.utils.serializeTransaction(txn as ethers.UnsignedTransaction);
+			return wei(await OptimismGasPriceOracleContract.getL1Fee(serializedTxn));
+		} catch (e) {
+			return null;
```
fritzschoff, line 64: console.error it?

fritzschoff: or pass it to the handle error func?

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -3,73 +3,54 @@ import { useQuery, UseQueryOptions } from 'react-query';
 import { QueryContext } from '../../context';
 import { GasPrices } from '../../types';
 import { formatGwei } from '../../utils';
+import Wei, { wei } from '@synthetixio/wei';
```
fritzschoff, line 6: `Wei` doesn't seems to be used?

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -3,73 +3,54 @@ import { useQuery, UseQueryOptions } from 'react-query';
...
-	fastWait: number;
-	fastest: number;
-	fastestWait: number;
-	gasPriceRange: Record<number, number>;
-	safeLow: number;
-	safeLowWait: number;
-	speed: number;
+export const computeGasFee = (baseFeePerGas: ethers.BigNumber, maxPriorityFeePerGas: number) => {
+	return {
+		maxPriorityFeePerGas: wei(baseFeePerGas, 9).toBN(),
```
fritzschoff, line 15: isn't maxPriorityFeePerGas the tip we are sending the miner? 
```
When you submit a transaction you will also provide a "tip" to the miner. This is the maxPriorityFeePerGas
 field. The bare minimum you should tip the miner is 1 wei, otherwise there is no incentive for 
the miner to #execute the transaction. The higher your tip, the more likely your transaction will be included in the block. 
This concept is pretty much the same as pre-London fork.
```

fritzschoff: Feel like the second param of the func should be mapped to the same key as in the return object

bachstatter: Oh I think this is a good catch. `maxPriorityFeePerGas` should be `maxPriorityFeePerGas` 😅 (ie. second arg to compute gas)?
@0xclem You agree with this, or is me and max misunderstanding something?

I do remember thinking the gas estimates on L1 looked high...

packages/queries/src/mutations/useEVMTxn.ts
bachstatter: Addressed this in a new commit. If we're on optimism and we fail to get l1-fees we want to display an error

packages/queries/src/queries/network/useEthGasPriceQuery.ts
bachstatter: Addressed. Removed Wei and a few other unused variables

packages/queries/src/queries/network/useEthGasPriceQuery.ts
bachstatter: I fixed this and added a test for `comuteGas`

packages/queries/src/mutations/useEVMTxn.ts
bachstatter: This is the only thing I didn't address in my commits. I dont see any big reason for why we changed it, maybe @0xclem ? Had some reasoning?

0xclem: @bachstatter I explained it to you the other day. Because `gasLimit` is technically not a `wei` value. It's a decimal number wrapper in a `BN`. Now feel like putting it back if you want but you need to make sure it's in the correct format. We used to do `wei(gasLimit)` which is wrong. `wei(gaslimit, 0)` is probably the correct format. But it makes no difference with keeping it in a `BN`. So your call, happy with whatever you guys decide.

packages/queries/src/queries/network/useEthGasPriceQuery.ts
```diff
@@ -1,75 +1,56 @@
...
-	fastWait: number;
-	fastest: number;
-	fastestWait: number;
-	gasPriceRange: Record<number, number>;
-	safeLow: number;
-	safeLowWait: number;
-	speed: number;
+export const computeGasFee = (baseFeePerGas: ethers.BigNumber, maxPriorityFeePerGas: number) => {
+	return {
+		maxPriorityFeePerGas: wei(maxPriorityFeePerGas, 9).toBN(),
```
0xclem, line 13: yes good one, it was actually a mistake!

PR #123, opened 2021-11-18
fix/temporarily update the ovm mainnet exchanges subgraph endpoint

kmeraz:


dbeal-eth (approved):
the bug is with the hosted service, it won't let us deploy to the exchanger optimism url after the regenesis.

temp until we get the new subgraphs out.

PR #125, opened 2021-11-18
feat: serialize wei as string instead of hex

JChiaramonte7:
```
// Example serialization for wei(100)
"100.000000000000000000"
// Example serialization for wei(100, 0, true)
"100"
```


dbeal-eth (approved):
amazing change, thanks so much.

PR #128, opened 2021-11-29
Debt Timeseries Query

noahlitvin:


.github/workflows/main.yml
```diff
@@ -29,4 +29,4 @@ jobs:
       - run: npx lerna exec npm version 0.0.0-$(git rev-parse --short origin/${{ github.head_ref }})
       - run: npx lerna exec npm publish -- --tag dev
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
dbeal-eth, line 32: you might want to re-checkout this file since nothing fundamentally changes here and its not relevant to the PR

packages/contracts-interface/package.json
```diff
@@ -5,7 +5,7 @@
 	"source": "./src/index.ts",
 	"main": "./build/node/src/index.js",
 	"module": "./build/node/src/index.js",
-	"browser": "./build/index.js",
+	"browser": "./build/node/src/index.js",
```
dbeal-eth, line 8: :pray:

packages/queries/codegen.js
```diff
@@ -45,3 +45,12 @@ for (const f of fs.readdirSync('subgraphs')) {
 
 	fs.writeFileSync(`generated/${f.substr(0, f.length - 5)}SubgraphQueries.ts`, text);
 }
+
+for (const f of fs.readdirSync('subgraphs')) {
+	const text = cgt.gen({
+		schema: JSON.parse(fs.readFileSync('subgraphs/' + f)),
+		method: 'plain',
```
dbeal-eth, line 52: hmm, these aren't strictly queries, but I can see why we might want to include this here anyway.

packages/queries/src/index.ts
```diff
@@ -114,6 +116,14 @@ export function createQueryContext({
 		);
 	}
 
+	modFuncs.main = {};
```
dbeal-eth, line 119: cna you rename this to `subgraph` or something? when we fully fork over to these main subgraphs, we wont need to refer to it as main anymore from an outside perspective

packages/queries/src/queries/debt/useGetDebtTimeseries.ts
```diff
@@ -0,0 +1,93 @@
...
+		async () => {
+			const debtStates = await getDebtStates(
+				ctx.subgraphEndpoints.main,
+				{ first: 100000, orderBy: 'timestamp', orderDirection: 'asc' },
+				{ timestamp: true, debtRatio: true, totalIssuedSynths: true }
+			);
+			const debtSnapshots = await getDebtSnapshots(
+				ctx.subgraphEndpoints.main,
+				{
+					first: 100000,
```
dbeal-eth, line 30: you probably want to cap this to a more practical number now that we know, or have it be a parameter for the `useGetDebtTimeseries` function

packages/queries/src/queries/debt/useGetDebtTimeseries.ts
```diff
@@ -0,0 +1,93 @@
...
+							),
+						});
+					}
+				});
+			}
+
+			return timeseries
+				.filter((d) => d.timestamp > 0)
+				.sort(function (a, b) {
+					return a.timestamp - b.timestamp;
```
dbeal-eth, line 83: sort normally shuoldn't be necessary here with just one address, right?

dbeal-eth: also this could be more concise using `_.sortBy(timeseries, 'timestamp')`, if you are ok with using a sprinkle of lodash

dbeal-eth (approved):
just fix the few things. thanks for implementing this!

packages/queries/codegen.js
noahlitvin: Yeah, I needed this because we can't call hooks within hooks.

packages/queries/src/queries/debt/useGetDebtTimeseries.ts
noahlitvin: Sorting seemed necessary from testing the output, but I didn't dig into why tbh. Happy to use `sortBy` here.

PR #129, opened 2021-12-02
bumped ethers to latest version

fritzschoff:
With that version, ethers supports the optimism network with the infura provider class


bachstatter:
CI failing, might need some lerna fiddeling? I dont have too much experience so might be worth asking someone with lerna experience. If we dont have a clear answer I'm happy to help figuring it out

fritzschoff:
@bachstatter seems like @noahlitvin  is having the same issue. He seems to find a workaround in his #130 PR but that doesn't seems to do the trick for me...

fritzschoff:
@bachstatter no idea what is going on 😆 . If you have an idea, happy to hear it

PR #133, opened 2021-12-07
ethers v5.5.2

fritzschoff:


fritzschoff:
Hey @dbeal-eth , it seems there is an issue with your generated output in the generated queryFuncs.ts file. `The inferred type of 'default' cannot be named without a reference to 'ethers/node_modules/@ethersproject/abstract-provider'. This is likely not portable. A type annotation is necessary.` Do you got an idea for an fix?

PR #137, opened 2021-12-15
update snapshot getScores queries

andytcf:


packages/queries/src/queries/gov/useProposalQuery.ts
```diff
@@ -142,46 +96,48 @@ const useProposalQuery = (
...
+					vote.scores = space.strategies.map(
+						(_: SpaceStrategy, key: number) => vote.vp_by_strategy[key] || 0
+					);
+					vote.balance = vote.vp;
+					return vote;
+				});
+			} else {
+				const block = parseInt(proposal.snapshot);
+
+				const [scores] = await Promise.all([
```
bachstatter, line 117: Can remove Promise.all since it's only one promise

packages/queries/src/queries/gov/useProposalQuery.ts
```diff
@@ -142,46 +96,48 @@ const useProposalQuery = (
...
-						vote.balance = vote.scores.reduce((a: number, b: number) => a + b, 0);
-						vote.profile = profiles[getAddress(vote.voter)];
-						return vote;
-					})
-					.filter((vote) => vote.balance > 0)
-					.sort((a, b) => b.balance - a.balance),
-				(a) => getAddress(a.voter)
-			);
+			if (proposal.state === 'closed') {
+				mappedVotes.map((vote) => {
```
bachstatter, line 107: I would not use `.map` but `.foreach` so we communicate that this loop mutate stuff

packages/queries/src/queries/gov/useProposalQuery.ts
```diff
@@ -7,62 +7,9 @@ import { SpaceData, Vote, SpaceStrategy, Proposal, ProposalResults } from '../..
...
-				});
-				resolve(ensNames);
-			})
-			.catch((error: any) => {
-				reject(error);
-			});
-	});
-}
-
-export async function getProfiles(addresses: any) {
```
bachstatter, line 50: Are we happy with not showing ens?

bachstatter (approved):
Left some comments, could probably make the code a bit more maintainable. But nothing really major, especially considering we really want to get this out before the election. I guess the only thing would be the ens stuff, isn't it important to show the ens? 

If you're confident it works we can go ahead with it and then I'm happy to do cleaning after the elections

PR #143, opened 2022-01-11
add handleSwitchChain func

fritzschoff:
Clem did that for Aelin and was encouraging us to do the same. I thought that it would be a good addition to the provider package.
Code used in Aelin
```
const handleSwitchChain = useCallback(async () => {
        if (!provider || !network?.id || !walletAddress) return;
        const web3Provider = provider as ethers.providers.Web3Provider;
        if (!web3Provider.provider || !web3Provider.provider.request) return;
        const newNetworkId = getCorrespondingNetwork(network?.id, isOVM);
        const formattedChainId = ethers.utils.hexStripZeros(BigNumber.from(newNetworkId).toHexString());
        await web3Provider.provider.request({
            method: 'wallet_switchEthereumChain',
            params: [{ chainId: formattedChainId }],
        });
    }, [isOVM, provider, network?.id, walletAddress]);
```

⚠️ ⚠️ ⚠️
Made some test and it seems only the Coinbase wallet and MetaMask are supporting that request. Tried with the trust wallet and nothing happened in the app

### DIFF
--- a/packages/providers/src/types.ts
+++ b/packages/providers/src/types.ts
@@ -13,15 +13,3 @@ type L1Provider = ethers.providers.Web3Provider | ethers.providers.InfuraProvide
 type L2Provider = OvmProvider;
 
 export type SynthetixProvider = L1Provider | L2Provider;
-
-export enum NetworkId {
-	Mainnet = 1,
-	Ropsten = 3,
-	Rinkeby = 4,
-	Goerli = 5,
-	Kovan = 42,
-	'Mainnet-Ovm' = 10,
-	'Kovan-Ovm' = 69,
-	'Local-Ovm' = 420,
-	'Local' = 31337,
-}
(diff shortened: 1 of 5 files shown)


packages/providers/src/types.ts
```diff
@@ -13,3 +13,15 @@ type L1Provider = ethers.providers.Web3Provider | ethers.providers.InfuraProvide
 type L2Provider = OvmProvider;
 
 export type SynthetixProvider = L1Provider | L2Provider;
+
+export enum NetworkId {
```
fritzschoff, line 17: would reuse the enum from the contracts-interface module, but if I import that module in the provider module, lerna is not symlinking the dependency and the build is crashing!

fritzschoff: Error is: can't resolve @synthetixio/contracts-interface

bachstatter:
Since they are different packages if you want to use something you need to install that package as a dependency.
Added the contract-interface as deps so we have access to `NetworkId`

bachstatter (approved):

bachstatter:
Before merging you can install this version and make sure it behaves as you expect: `npm i @synthetixio/providers@0.0.0-5019e64`

PR #159, opened 2022-02-01
updated subgraph url

fritzschoff:
In staking we only care about lockedValue from `useGlobalStakingInfoQuery`. I depricated that hook and added a simpler `useLockedSnxQuery`.

Also upgrading `codegen-graph-ts`

### DIFF
--- a/packages/queries/src/queries/staking/useGlobalStakingInfoQuery.ts
+++ b/packages/queries/src/queries/staking/useGlobalStakingInfoQuery.ts
@@ -10,11 +10,19 @@ const useGlobalStakingInfoQuery = (
 	ctx: QueryContext,
 	options?: UseQueryOptions<GlobalStakingInfo>
 ) => {
-	const snxHoldersQuery = useGetSNXHolders(ctx.subgraphEndpoints.subgraph, {
-		first: 1000,
-		orderBy: 'collateral',
-		orderDirection: 'desc',
-	});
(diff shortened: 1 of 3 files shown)


packages/queries/src/queries/staking/useGlobalStakingInfoQuery.ts
```diff
@@ -37,7 +45,7 @@ const useGlobalStakingInfoQuery = (
 				ctx.snxjs!.contracts.SystemSettings.issuanceRatio(),
 			]);
 
-			const lastDebtLedgerEntry = wei(unformattedLastDebtLedgerEntry, 27);
+			const lastDebtLedgerEntry = wei(unformattedLastDebtLedgerEntry);
```
bachstatter, line 48: I asked some questions about this change in discord. This change in combination with patching `useGetSNXHolders` to use `snxholders` instead of `sNXHolder` seems to fix the calculation.
But I dont understand how/why it worked with `27` precision before.
PR: https://github.com/Synthetixio/staking/pull/904
Deploy preview:
https://staking-git-fix-apy-calc-synthetixio.vercel.app/

bachstatter:
Would be great to get a review from you @dbeal-eth

packages/queries/src/queries/staking/useGlobalStakingInfoQuery.ts
bachstatter: Ignore this..

PR #174, opened 2022-03-18
updated load provider function

fritzschoff:


bachstatter (commented):
I prefer to just use the project id and whitelisting domains. The project secret should only be used in backend services. If we add it to the frontend anyone can grab the secret and use it on their domain

fritzschoff:
yeah good point. I'm using loadprovider function in a SSR func but better doing it with ethers js directly

PR #184, opened 2022-04-01
Using NPM workspace now

### LINKED ISSUE #183: Move away from Lerna
Lerna is not supported anymore and last PR that was merges was 14 months ago.
My idea is to go with npm > 8 workspaces or NX

fritzschoff:
close #183


fritzschoff:
@dev dont merge, need to update the updateDependecy.yml

bachstatter (approved):
I'm all for switching out lerna. I feels like a little scary change though, and I dont have experience with workspaces. We need to make sure we test this a lot.

Lets catchup tonight before merging

bachstatter (changes requested):
For some reason the synthetix version used when consuming this packages are wrong.. Need to get to the bottom of that

PR #185, opened 2022-04-07
removed lerna

### LINKED ISSUE #183: Move away from Lerna

fritzschoff:
close #183


.github/workflows/main.yml
```diff
@@ -13,22 +13,51 @@ jobs:
       # Setup .npmrc file to publish to npm
       - uses: actions/setup-node@v2
         with:
-          node-version: '14.x'
+          node-version: '16.x'
```
bachstatter, line 16: I think we should use `ghcr.io/synthetixio/docker-node/alpine:16.14`,
Staking uses that container (although still on 14)
https://github.com/Synthetixio/staking/blob/dev/.github/workflows/audit_build_verify.yml#L37

You can see available versions here:
https://github.com/orgs/Synthetixio/packages/container/package/docker-node%2Falpine

This docker container comes with `yarn` installed so we can skip `npm i -g yarn` below.

.github/workflows/main.yml
```diff
@@ -13,22 +13,51 @@ jobs:
...
           registry-url: 'https://registry.npmjs.org'
-      - run: npm ci
-      - run: npx lerna bootstrap
-      - run: npm run lint
-      - run: npx lerna run build
-      - run: npx lerna run test
+      - run: npm i -g yarn
+      - run: yarn install
+      - run: yarn run build
+      - run: yarn run lint:fix
```
bachstatter, line 21: Probably dont want to fix in CI. We like the code we commit to be the same in the package. Should be up to the dev to fix before hand

.github/workflows/main.yml
```diff
@@ -13,22 +13,51 @@ jobs:
...
+      - run: yarn run test
```
bachstatter, line 22: Will this run the test for all components? I think it should be `yarn workspaces run test`
(https://classic.yarnpkg.com/en/docs/cli/workspaces#toc-yarn-workspaces-run)

.github/workflows/main.yml
```diff
@@ -13,22 +13,51 @@ jobs:
...
       - run: git fetch origin ${{ github.head_ref }}
       - run: echo $(git rev-parse --short origin/${{ github.head_ref }})
-      # Remove the next line and uncomment the two below to fix dev build issues from inter-package version mismatch issues.
-      - run: npx lerna exec npm version 0.0.0-$(git rev-parse --short origin/${{ github.head_ref }})
-      # - run: git checkout -b temp
-      # - run: npx lerna version 0.0.0-$(git rev-parse --short origin/${{ github.head_ref }}) --exact --no-push -y
-      - run: npx lerna exec npm publish -- --tag dev
+
+      - run: |
+          npm -w contracts-interface version 0.0.0-$(git rev-parse --short origin/${{ github.head_ref }})
```
bachstatter, line 30: Should we be using yarn here? `yarn workspace version 0.0.0-$(git rev-parse --short origin/${{ github.head_ref }})`

.github/workflows/updateDependency.yml
```diff
@@ -28,27 +25,46 @@ jobs:
...
-      - name: Publish packages
-        run: lerna publish from-package --yes
+          npm -w contracts-interface version ${{ env.new_version }}
+          npm -w optimism-networks version ${{ env.new_version }}
+          npm -w providers version ${{ env.new_version }}
+          npm -w queries version ${{ env.new_version }}
+          npm -w transaction-notifier version ${{ env.new_version }}
+          npm -w wei version ${{ env.new_version }}
+
+      - run: pm -w contracts-interface publish --tag latest
```
bachstatter, line 44: typo here `pm` -> `npm`. But as said above I think we should be using yarn

package.json
```diff
@@ -1,64 +1,64 @@
...
-		"bootstrap": "^5.1.0"
+	"homepage": "https://github.com/Synthetixio/js-monorepo#readme",
+	"workspaces": [
+		"contracts-interface",
+		"transaction-notifier",
+		"optimism-networks",
+		"wei",
+		"providers",
+		"queries"
+	],
```
bachstatter, line 32: This is a slightly annoying comment. But to me it make sense to keep the `packages` folder. 
and have the workspace be `"workspaces": [ "packages/*]`.
Two reasons: 
1. That's usually what I see in other monorepos.
2. If we add a new package it will be picked up automatically

package.json
```diff
@@ -1,64 +1,64 @@
...
+		"eslint-plugin-prettier": "^4.0.0",
+		"husky": "^7.0.4",
+		"jest": "^27.5.1",
+		"prettier": "^2.6.2",
+		"react": "^17.0.2",
+		"react-dom": "^17.0.2",
+		"ts-loader": "^9.2.8",
+		"typescript": "^4.6.3",
+		"webpack": "^5.71.0",
+		"webpack-cli": "^4.9.2"
```
bachstatter, line 62: Do we need the dependencies here? I thought yarn workspaces would take care of that? 
Or does running `yarn install` produce these dependencies automatically?

bachstatter (changes requested):
Left some comments. Can have a chat next week

.github/workflows/main.yml
fritzschoff: yes it runs for all the workspaces

package.json
fritzschoff: doesn't that include node_modules and .github dirs?

PR #197, opened 2022-05-31
Limit files published to NPM in wei and add src to the published packages

noisekit:
Looks like wei was missing `files` limits on what we published so it shipped with a bunch of dev-only files

image

Additionally I have added `src` too. We do reference `source` in package.json but do not ship sources.

Generally shipping sources is good, because it gives developers opportunity to adjust bundling and build from our own sources, optimising overall bundle size of the end app.

After the update wei has only necessary files

image


bachstatter (approved):
Nice!

PR #198, opened 2022-05-31
Remove unused bootstrap dependency

noisekit:
I suspect it was added by accident like `npm i lerna bootstrap`
Will follow up with more automated way to check for depencencies (with `depcheck`) next


bachstatter (approved):
🕵️

PR #200, opened 2022-06-01
Ensure we can publish on repeated rebuilds

noisekit:
If we re-run the same workflow again, because we rely on the same git SHA version we cannot publish new version on NPM.

The fix: append release version with `.${{ github.run_attempt }}` in case it is greater than 1. So for all the one-time runs we still have same 8-char commit sha version like `0.0.0-c69c1461` but in case we manually rerun the workflow we will have versions like `0.0.0-c69c1461.2`

Alternative solution would be to skip publishing entirely on run_attempt > 1 but in that case we may not be able to publish something that could not be published on the initial run (in case of network error or smth). So republishing with extra number seems more robust.

Running smooth:
image

And the repeated build worked as expected too:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,10 +21,19 @@ jobs:
       - run: npm run test
 
       # publish
-      - run: git config --global user.email "ci@cc.snxdao.io" && git config --global user.name "Synthetix CI"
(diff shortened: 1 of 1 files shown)


bachstatter (approved):
Splendid

PR #202, opened 2022-06-02
Yarn Berry instead of NPM workspaces

noisekit:
- [x] Update DECISIONS.md with why Yarn case
- [x] Replace NPM with Yarn and add all necessary Yarn configs and plugins
- [x] Ensure `main` github workflow runs
- [x] Ensure `updateDependency` github workflow runs

## `main` workflow

image

image

## `updateDependency` workflow
Cannot be fully tested, but separate "non-destructive" parts of it work well

Updating synthetix can be performed over whole monorepo with `yarn up`!

image

It updates relevant package.json and yarn.lock

image

At this point we will publish all versions at the same time, but later we can rely on `yarn version` to only update what depends on `contracts-interface` and `synthetix` finally breaking us out of symmetric publishing of all packages

Running workflow without pushing/publishing works as expected. All packages got their new versions and would be published + changes git tagged and pushed

image

## Versioning

Yarn is properly replacing all the versions when publishing packages

image


0xjocke (approved):
Very clean!

fritzschoff:
🍓

PR #204, opened 2022-06-03
Automate dependency checks with CI

noisekit:
- [x] add yarn config option `defaultSemverRangePrefix` to always use exact version (instead of a default `^`)
- [x] `deps` script to walk over deps and find missing or extra deps
- [x] `deps-mismatch` script to find mismatched versions between workspace packages
- [x] `deps-circular` script to fins circular references (no matter of depth) between workspace packages (aka `x -> y -> z -> x`)
- [x] add deps checking to `main` GH action
- [x] fix some loose minor issues with deps
- [x] for the time being just add a list of dev deps that we use for building and checks to the ignore list. May revise later and fix it in a different way

## results

Passing in CI and locally

image


0xjocke (approved):
Didn't dig deep into all the logic but it all looks good to me

PR #205, opened 2022-06-08
Use synths balances query type fix

0xjocke:
The `BalanceMap` will only have synths keys for synths that have a balance greater than 0.

Previously typescript would not complain if you did:
`useSynthsBalancesQuery().data?.balancesMap.sUSD.balance`
But if your balance is 0 this would be a runtime error.

The correct usage to get the balance. and now enforced by typescript:
`useSynthsBalancesQuery().data?.balancesMap.sUSD?.balance`

I also made some other improvement, minimising `as` and `!` and correct return type for `SynthUtil.synthsBalances`


noisekit (approved):

bachstatter:
Tried it out on staking, and ran `yarn tsc` without any problems.  I fixed the only incorrect usage earlier: https://github.com/Synthetixio/staking/pull/1103

I dont think we need to make a release for this as there's a smart contract release later today/ tonight

PR #207, opened 2022-06-08
Unlock dependencies versions

noisekit:
Moving away from locking dependendcy versions as recently discussed @dbeal-eth and @drptbl. See https://discordapp.com/channels/413890591840272394/981061962572124181 for more context

Couple of notes and workarounds:
- [x] `synthetix` version is still exact and locked
- [x] `react-query` version is temporarily locked as any newer version breaks types and `tsc` build fails (this needs to be fixed on codegen level separately)


0xjocke (approved):
Nice!

PR #209, opened 2022-06-09
Include codegen-graph-ts into monorepo

noisekit:
Because `codegen-graph-ts` depends on `wei` and `queries` depend on `codegen-graph-ts` we ended up having multiple `wei` instances of different versions in the lockfile

Notable changes:
- [x] Rename to `@synthetixio/codegen-graph-ts` so we can publish (and have publishing rights for synthetix npm user)
- [x] Prevent package.json from being pulled into the build

Monorepo updates:
- [x] Switch to --topological-dev to include dev deps in building order
- [x] Document adding external library into the monorepo preserving git history

Some other codegen-graph-ts updates include:
- [x] Fix deps issues and mismatches
- [x] Align TS and testing configs
- [x] Apply Prettier and Eslint rules to the code

### 1. Prepare original repo

- Create a separate branch `move-to-monorepo` 
- Create the intended destination folder inside monorepo `mkdir -p tools/codegen-graph-ts`
- Move all the package files into `tools/codegen-graph-ts`
- Remove all the files that won't be used (CI config, lockfile, etc)

### 2. Add remote to monorepo

```sh
cd ~/synthetix/js-monorepo
git remote add codegen-graph-ts ~/synthetix/codegen-graph-ts
git fetch --all
```

### 3. Merge original repo branch
Using `--allow-unrelated-histories` allows merging independent git history

```sh
git merge codegen-graph-ts/move-to-monorepo --allow-unrelated-histories
```

Because we moved all the files into the separate folder we have no merge conflicts and at the same time we have full history added to the git tree

Now we can remove remote as it is no longer necessary and cleanup all the added tags too

```sh
git remote remove codegen-graph-ts

# Cleanup all local tags and re-fetch existing tags without just removed `codegen-graph-ts` remote 
git tag -l | xargs git tag -d
git fetch --tags
```


tools/codegen-graph-ts/package.json
```diff
@@ -0,0 +1,61 @@
...
+		"commander": "^8.1.0",
+		"graphql": "15.5.0",
+		"graphql-request": "3.4.0",
+		"lodash": "^4.17.21"
+	},
+	"devDependencies": {
+		"@types/commander": "^2.12.2",
+		"@types/lodash": "4.14.169",
+		"eval": "^0.1.6",
+		"node-fetch": "cjs",
```
bachstatter, line 40: version `cjs`?

0xjocke (approved):
Nice! I stepped through your commits and it makes sense to me

PR #211, opened 2022-06-09
Add useGetElectionsPeriodStatus

bachstatter:


packages/queries/src/queries/gov/constants.ts
```diff
@@ -1,3 +1,8 @@
+import { Ambassador } from '../../contracts/Ambassador';
+import { Grants } from '../../contracts/Grants';
```
jmzwar, line 2: Minor but this could all be 1 line
import { Ambassador, Grants, ...} from '../../contracts'

jmzwar (approved):

jmzwar (approved):

PR #214, opened 2022-06-14
Independent versioning

noisekit:
The only place we need to change the behaviour is our auto-update github action

Notable differences:
1. `monorepo_version` no longer exists
2. `synthetix_version` will only update contracts package (`root` will not be updated as it does not depend on contracts and no other packages are updated for the same reason)
3. separate github actions workflow added to publish manually one of the packages in monorepo (expects package name and version)
4. dapps updates are a bit safer with `jq` plus in separate steps and verbose logging
5. there is no command in yarn to _automatically_ increase version on all updated packages - so we have a bit of bash for that
6. extra `deps:version` script added. without arguments it will print all cascading dependency updates for every monorepo package. When supplied with package name and version, it will update package to specified version (deferred!) and all of packages depending on it (as `patch`, also deferred!). Then we (or CI) can run `yarn version apply --all` to actually apply all the versions to packages.

## Running `yarn deps:version` without arguments
Screen Shot 2022-06-29 at 12 53 28

## Running `yarn deps:version @synthetixio/contracts-interface patch` with package and version
Screen Shot 2022-06-29 at 12 54 21

## Then after `yarn version apply --all` we can observe this git diff
Screen Shot 2022-06-29 at 12 55 13


0xjocke (approved):
Nice! Looks good to me. I think adding some documentation about how the two workflows works and should be used would be good

README.md
```diff
@@ -32,27 +34,71 @@ yarn build
...
+### Publishing new version after `synthetix` upgrade
 
 We have a GitHub workflow for publishing releases.
+
 To publish:
 
 1. Go here https://github.com/Synthetixio/js-monorepo/actions/workflows/updateDependency.yml
 2. Click Run Workflow
+3. Fill `synthetix_version` in a format of `1.2.3` for the version
+4. Run the workflow
```
jmzwar, line 50: Nice

jmzwar (approved):

PR #218, opened 2022-06-21
Remove deprecated @eth-optimism/provider dependency and replace with standard types

noisekit:
Apparently `@eth-optimism/provider` is deprecated long time ago. Optimism uses standard RPC provider type interface now

- [x] Will test `dev` version on the `staking` first to ensure it is not breaking and types are compatible
- [x] Test `dev` in yarn v3 upgrade branch to confirm audit is working

Audit works now!
image

I did quite a bit of testing on L1 at https://staking-git-test-dev-synthetixio.vercel.app (staking, minting, burning, etc). And found no issues at this point at all.

As an additional bonus we lost another ~200kb of page load bundle size!

The PR is here: https://github.com/Synthetixio/staking/pull/1143
The only thing I needed to remove one manual type declaration, so it is not entirely "non-breaking" (it looks like TS makes any change - breaking change, but seems like nobody bothers publishing major versions for such things. Semver is killed by TS 🤦)

PS: unblocks https://github.com/Synthetixio/staking/pull/1100

### DIFF
--- a/packages/providers/src/types.ts
+++ b/packages/providers/src/types.ts
@@ -1,7 +1,6 @@
 import { ethers } from 'ethers';
-import { OptimismProvider } from '@eth-optimism/provider';
 
-export type OvmProvider = typeof OptimismProvider;
+export type OvmProvider = ethers.providers.JsonRpcProvider;
 
 export type ProviderConfig = {
 	networkId?: ethers.providers.Networkish;
--- a/packages/providers/src/types.ts
+++ b/packages/providers/src/types.ts
@@ -1,24 +1,9 @@
 import type {
-	JsonRpcProvider,
-	Web3Provider,
+	ExternalProvider,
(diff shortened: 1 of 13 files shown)


packages/providers/src/types.ts
```diff
@@ -1,15 +1,32 @@
...
+	JsonRpcProvider,
+	Web3Provider,
+	InfuraProvider,
+	Networkish,
+	ExternalProvider,
+	JsonRpcSigner,
+} from '@ethersproject/providers';
 
-export type OvmProvider = typeof OptimismProvider;
+export declare class OptimismProvider extends JsonRpcProvider {
```
bachstatter, line 10: I wonder if we actually need a different custom type `OptimismProvider`

0xjocke (approved):

PR #221, opened 2022-06-27
Fix notifying apps

0xjocke:
I think the update dependency broke when we stopped listening to versions in the apps being notified.
This run:
https://github.com/Synthetixio/js-monorepo/runs/7031864303?check_suite_focus=true
Never made it to staking:
https://github.com/Synthetixio/staking/actions/workflows/updateDependency.yml

I also stopped notifying Kwenta's repos since they dont seem to rely on it for upgrading

### DIFF
--- a/.github/workflows/updateDependency.yml
+++ b/.github/workflows/updateDependency.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Notify dApps
         run: |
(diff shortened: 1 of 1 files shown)


noisekit (approved):
Yeah if that param was removed then we should have updated curls too 😅

PR #229, opened 2022-07-19
@synthetixio/contracts

noisekit:
- [x] deployments.json exploded
- [x] ts files generated for deployments
- [x] tsc compilation added
- [ ] ~generate ts files for all other jsons~. UPDATE: I think this can be done as we go depending on what we actually are using. So prob not as part of this PR

The results of code generation

# Sources:
image

# Targets:
image

# Synths:
image

# Used in `staking`, no compile problems and things run fine
image


jmzwar (approved):

0xjocke (approved):

bachstatter:
Hmm one other thing I thought about. I've had problems using enum in npm librararies with typescript before. 
Maybe the Synths can just as an object?

PR #243, opened 2022-08-10
Import UI

noisekit:
- Import UI repo into the package (`v4` branch)
- Adjust build and code to match monorepo standards


v3/theme/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "@rollup/plugin-node-resolve": "^13.3.0",
+    "@storybook/addon-essentials": "^6.5.7",
+    "@storybook/addon-interactions": "^6.5.7",
+    "@storybook/addon-links": "^6.5.7",
+    "@storybook/builder-webpack5": "^6.5.7",
+    "@storybook/manager-webpack5": "^6.5.10",
+    "@storybook/react": "^6.5.7",
+    "rollup": "^2.77.2",
+    "rollup-plugin-babel": "^4.4.0",
+    "rollup-plugin-peer-deps-external": "^2.2.4",
```
fritzschoff, line 44: any reason why we are using rollup here and in the rest of the monorepo webpack?

0xjocke (approved):
Looks good to me. Maybe worth having a discussion if rollup adds an value as Max said

v3/theme/package.json
sunnyvempati: we can remove it; this was very quickly thrown together.  i can push up a fix in a subsequent pr to use babel instead

sunnyvempati (approved):

PR #245, opened 2022-08-10
Remove Rollup from v3-theme

noisekit:
Switch to our standard tsc build for libs as we do it in all other packages.
There was no specific reason to use rollup so we can do it.

# Storybook OK

image

# Build OK

image


sunnyvempati (approved):
lgtm 👍🏽

PR #251, opened 2022-08-16
Update eslint

noisekit:
We had warnings about eslint incompatibilities with typescript
Also missing react/react-hooks (will be needed for UI apps, including v3/ui)

Fixed ESlint errors for v3/ui

Extracting this from staking import future PR to reduce number of changes


.eslintrc.js
```diff
@@ -0,0 +1,79 @@
...
+        exports: 'always-multiline',
+        functions: 'never',
+      },
+    ],
+    indent: 'off',
+    quotes: ['error', 'single'],
+    'no-undef': 'error',
+    'prefer-const': 'error',
+    semi: ['error', 'always'],
+    'no-console': 'error',
```
jmzwar, line 27: This is probably a good idea but bear in mind when we move staking here - there are a few places where we a logging in try/catches

.eslintrc.js
```diff
@@ -0,0 +1,79 @@
...
+  rules: {
+    'no-mixed-spaces-and-tabs': ['warn', 'smart-tabs'],
+    'comma-dangle': [
+      'error',
+      {
+        arrays: 'always-multiline',
+        objects: 'always-multiline',
+        imports: 'always-multiline',
+        exports: 'always-multiline',
+        functions: 'never',
```
jmzwar, line 19: No clashes here with prettier?

jmzwar (approved):
Looks good. Will be good to have this aligned across everything.

PR #261, opened 2022-08-22
Separate subgraph query runtime dependency from dev-only codegen tool

noisekit:
Currently we bundle up together dev-only codegen with runtime depdendency to generate grapql query. This requires from us to use deep-linked files so we do not accidentally import non-browser code that codegen uses.

Splitting them in two separate libs is the easy solution.

This will become even more explicit when we add to git and commit autogenerated code.

PS: extracted from a bigger refactor PR


tools/codegen-graph-ts/src/lib/axios.ts
```diff
@@ -1,4 +0,0 @@
-import axios from 'axios';
-
-// this is basically just a re-export which can be used by the genned code
-export default axios;
```
bachstatter, line 4: A little confused why this was needed before

0xjocke (approved):

PR #262, opened 2022-08-22
Commit generated contracts

noisekit:
1. Refactor code generation for contracts and place them under `src`
2. Commit all the contracts in `src`
3. Include CI check for changed files after build in case we forgot to commit changes


0xjocke (approved):
Looks good. One slight improvement now that they live in src would be to add a comment on the top of each file saying some like this:
`/* This files is autogenerated and should not be be edited manually */`

bachstatter:
Nice!

PR #264, opened 2022-08-22
Refactor and commit generated code in queries

noisekit:
One more piece of the puzzle.
Note I have moved test files to be close to what they actually test (and renamed accordingly when name as not matching) and excluded test files from typescript build.

### DIFF
--- a/tools/codegen-graph-ts/src/gen/segments/body.ts
+++ b/tools/codegen-graph-ts/src/gen/segments/body.ts
@@ -78,7 +78,10 @@ export function multiBody(e: Entity) {
     let results: Pick<${e.name}Result, K>[] = [];
 
     do {
-        if (paginationKey && paginationValue) paginatedOptions.where![paginationKey] = paginationValue as any;
+        if (paginationKey && paginationValue) {
+          // @ts-ignore
+          paginatedOptions.where![paginationKey] = paginationValue as any;
+        }
 
         const res = await axios.post(url, {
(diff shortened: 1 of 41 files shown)


jmzwar (approved):

tools/codegen-graph-ts/src/gen/segments/body.ts
```diff
@@ -78,7 +78,10 @@ export function multiBody(e: Entity) {
     let results: Pick<${e.name}Result, K>[] = [];
 
     do {
-        if (paginationKey && paginationValue) paginatedOptions.where![paginationKey] = paginationValue as any;
+        if (paginationKey && paginationValue) {
+          // @ts-ignore
```
bachstatter, line 82: I dont think this is needed. I did a find replace on `// @ts-ignore` -> `''`
in `/js-monorepo/packages/queries/src/subgraph/` the `yarn tsc` was still happy. 
Could it be that TS was in some weird states when you were running in to problems with this?

0xjocke (changes requested):
Could you try removing the ts-ignore? If you still have problems we can leave it in but as said in previous comment worked fine for me

tools/codegen-graph-ts/src/gen/segments/body.ts
bachstatter: Does not work when staking in imported to monorepo 🤔😢

0xjocke (approved):

PR #266, opened 2022-08-23
Refactor codegen and include generated code

noisekit:
- Cleanup all the mentions of `generated` code
- Move tests close to tested files (renaming accordingly)

PS: blocked by #264


tools/codegen-graph-ts/tests/exchanges.ts
```diff
@@ -0,0 +1,2376 @@
+import Wei, { WeiSource, wei } from '@synthetixio/wei';
```
bachstatter, line 1: This file doesn't make sense to me, it's in a test file but looks like one of the generated subgraph files?  Was it accidentally committed?

0xjocke (approved):

PR #267, opened 2022-08-23
Typescript monorepo rework

noisekit:
- Refactor typescript compilation to work better with monorepo (reference projects without type duplications)
- Remove unnecessary `node/src` folders out of `dist` by using `rootDir` option
- Remove root tsconfig (we must only build packages one by one and never full monorepo)

PS: blocked by #266 and #264


packages/contracts-interface/package.json
```diff
@@ -6,16 +6,16 @@
...
+  "module": "./build/index.js",
+  "browser": "./build/index.js",
+  "types": "./build/index.d.ts",
   "files": [
     "build",
     "src"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "rm -rf build && tsc --build --force",
```
jmzwar, line 18: Why is the --force necessary?

0xjocke (approved):
I think everything looks good to me.

> Refactor typescript compilation to work better with monorepo (reference projects without type duplications)

Not sure I fully understand what the reference project setting does.  Do you think its worth adding a note in the readme about the monorepo setup?

> Remove root tsconfig (we must only build packages one by one and never full monorepo)

hmm what is the implication of this? It looks like the build task in the root `package.json`  runs `tsc`?

PR #268, opened 2022-08-24
theme file: button component

fritzschoff:


v3/theme/src/theme.ts
```diff
@@ -1,6 +1,23 @@
-import { extendTheme } from '@chakra-ui/react';
+import { ComponentStyleConfig, extendTheme } from '@chakra-ui/react';
+import { mode } from '@chakra-ui/theme-tools';
+
+const Button: ComponentStyleConfig = {
```
sunnyvempati, line 4: looks good; make sure you add to `extendTheme` below 👍🏽

fritzschoff: https://github.com/Synthetixio/js-monorepo/pull/268/files#diff-cef6cb3eac772efa2f9a7b2ce8f09caed02083f763225013ed5a867111c7ae5eR165

sunnyvempati (approved):
lgtm

v3/theme/src/theme.ts
```diff
@@ -1,6 +1,38 @@
...
+    focusBorderColor: 'cyan.500',
+    iconColor: 'cyan.500',
+  },
+};
+const NumberInput: ComponentStyleConfig = {
+  defaultProps: {
+    focusBorderColor: 'cyan.500',
+  },
+};
 export const theme = extendTheme({
```
bachstatter, line 34: @sunnyvempati  @fritzschoff  Sorry a little late to the game. But I reccomend not using `extendTheme` untill you actually pass it to the component. If you define it like this you get proper typing:
```
import { ChakraTheme } from '@chakra-ui/react';
export const theme: Partial<ChakraTheme> = {}
```
`extendTheme` is typed as any
unknown

PR #270, opened 2022-08-25
TSC Incremental builds with cleaning target dir

noisekit:
after deleting build dir next run will not produce any results because tsbuildinfo thinks files are already built and skips that.
Instead we can use `tsc`'s own `--clean` option.


0xjocke (commented):
Looks good to me but build is failing 🤔

0xjocke (approved):

PR #271, opened 2022-08-25
fix: v3 theme import issue

Rickk137:


fritzschoff:
@Rickk137 can you pull master and see if that works for you? Or try it out on master. Should work now

sunnyvempati:
this is fixed with nikita's PR

PR #273, opened 2022-08-26
Standalone builds

noisekit:
1. Use `--recursive` option so we rely on package dependencies rather than package order in `workspaces` list (can now simplify it)
2. Add `standalone-install` and `standalone-build` scripts to `v3/ui` to illustrate how we need to build single package/service within monorepo (this way it also recurses down through all the dependencies too)
3. Update GH workflow to publish a package with `focus`

### DIFF
--- a/.github/workflows/publishSinglePackage.yml
+++ b/.github/workflows/publishSinglePackage.yml
@@ -31,7 +31,7 @@ jobs:
         env:
           YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - run: yarn install --immutable
+      - run: yarn workspaces focus '${{ github.event.inputs.package }}'
 
       - name: Config git identity
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/publishSinglePackage.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         env:
           YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - run: yarn install --immutable
+      - run: yarn workspaces focus '${{ github.event.inputs.package }}'
```
bachstatter, line 34: Maybe this is common knowledge but I wouldn't mind a comment explaining what focus does

v3/contracts/package.json
```diff
@@ -49,7 +49,7 @@
     "clean": "shx rm -rf ./artifacts ./cache ./coverage ./src/types ./coverage.json",
     "commit": "git-cz",
     "compile": "cross-env TS_NODE_TRANSPILE_ONLY=true hardhat compile",
-    "test": "hardhat test",
+    "test": "echo Skip contract tests || hardhat test",
```
bachstatter, line 52: Why dont we want to run hardhat tests?

bachstatter: hehe ok sounds good to me.  Good for @sunnyvempati  to be aware maybe

.github/workflows/publishSinglePackage.yml
bachstatter: Yea fair enough

0xjocke (approved):

PR #274, opened 2022-08-26
Publishing

noisekit:
1. Default to `prerelease` strategy (safer)
2. Update cascade to `prerelease` or `patch` depending on the intended parent package version
3. As we operate on inversed dependency tree, it is necessary to install all dependencies and not focus workspace

Screen Shot 2022-08-26 at 13 10 23

Screen Shot 2022-08-26 at 13 10 31

```sh
yarn deps:version @synthetixio/wei prerelease
yarn version apply --all

~/syn/js
➜ git st
## publishing...origin/publishing
 M packages/queries/package.json
 M packages/wei/package.json
 M tools/codegen-graph-ts/package.json
 M tools/generate-subgraph-query/package.json
```

the diff has "pre-release" versions for all involved packages:

image

### DIFF
--- a/tools/deps/version.js
+++ b/tools/deps/version.js
@@ -49,10 +49,31 @@ async function run() {
     await exec(`yarn workspace ${name} version --deferred ${version}`);
 
     for await (const pkg of cascade[name]) {
-      console.log(
-        `${fgCyan}Bumping dependency ${fgGreen}${pkg}${fgCyan} to ${fgYellow}patch${fgReset}`
-      );
-      await exec(`yarn workspace ${pkg} version --deferred patch`);
+      switch (true) {
+        // When publishing exact version like `1.2.3`, update all dependent packages with `patch` strategy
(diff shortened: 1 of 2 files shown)


tools/deps/version.js
```diff
@@ -49,10 +49,31 @@ async function run() {
     await exec(`yarn workspace ${name} version --deferred ${version}`);
 
     for await (const pkg of cascade[name]) {
-      console.log(
-        `${fgCyan}Bumping dependency ${fgGreen}${pkg}${fgCyan} to ${fgYellow}patch${fgReset}`
-      );
-      await exec(`yarn workspace ${pkg} version --deferred patch`);
+      switch (true) {
```
bachstatter, line 52: I never used a `switch(true)`. It would be the same as having `if`s with break inside them?

bachstatter: hehe almost

0xjocke (approved):

PR #276, opened 2022-08-26
Feat/multicall refactor

Rickk137:


v3/ui/src/components/accounts/Stake/index.tsx
```diff
@@ -129,23 +122,34 @@ export default function Stake({
...
     if (!snxProxy) return [];
-    const createAccountCall: MulticallCall = [snxProxy.contract, 'createAccount', [newAccountId]];
+
+    const createAccountCall: MulticallCall[] = [
+      {
+        contract: snxProxy.contract,
+        functionName: 'createAccount',
+        callArgs: [newAccountId],
+      },
+    ];
```
sunnyvempati, line 132: yeah buddy ➕

v3/ui/src/components/accounts/Stake/index.tsx
```diff
@@ -169,38 +173,35 @@ export default function Stake({
...
-        refetchAccounts!({ cancelRefetch: Boolean(accountId) }),
-      ]);
+      await Promise.all([refetchAccounts!({ cancelRefetch: Boolean(accountId) })]);
       if (!Boolean(accountId)) {
-        // router.push({
-        //   pathname: `/accounts/${newAccountId}`,
-        //   query: router.query,
-        // });
-        navigate(`/accounts/${newAccountId}`);
+        navigate(`/accounts/${newAccountId}?chain=${chain?.network}`);
```
sunnyvempati, line 197: good catch;  it would be nice to protect devs from remembering to do this.  i generally don't like overloading 3rd party lib functions but could be worth it as long as we don't abuse it lol

v3/ui/src/hooks/useApprove.ts
```diff
@@ -0,0 +1,57 @@
...
+      refetchAllowance();
+      config?.onSuccess && config.onSuccess();
+    }
+  }, [config, refetchAllowance, sufficientAllowance, writeAsync]);
+
+  return {
+    isLoading,
+    approve,
+    refetchAllowance,
+    requireApproval: !sufficientAllowance,
```
sunnyvempati, line 55: thoughts on moving `sufficientAllowance` stuff into `useApproveCall`?   

my rationale: `useApprove`'s responsibility is to make the approve call and `useApproveCall` which is the orchestrator has the responsibility to call `useApprove` if insufficient allowance, prior to calling the multicall.

sunnyvempati (commented):
looks awesome man, good stuff.  just one comment 🤙🏽

v3/ui/src/hooks/useApprove.ts
Rickk137: well the way i was thinking is ``useApprove`` job is to return a function for approving
approve only makes a tx if there is not sufficientAllowance. in that way if you just ``await approve()`` you are sure that it will only goes to next line if there is sufficientAllowance OR user have approved.
``useApproveCall`` (maybe we can improve the naming) is a hook that receives an approval info and a function to call after that.
this function can be a multi-call OR anything else, so it's  a bit more general

PR #278, opened 2022-08-29
Multi-package releases

noisekit:
`deps:version` now accepts comma-separated list of packages. Unfortunately does not allow to enforce specific version per package yet, but that is smth to do later (if we need it)

```sh
yarn deps:version '@synthetixio/wei,@synthetixio/contracts' prerelease
```

It is still possible to use single package there, no problem
```sh
yarn deps:version @synthetixio/wei prerelease
```

And it will still resolve all the cascaded updates

## Few examples:

Multi-package updated to `minor`
```sh
yarn deps:version '@synthetixio/wei,@synthetixio/contracts' minor 
```
Results in:
image

Single package `prerelease` update
```sh
yarn deps:version '@synthetixio/wei,@synthetixio/contracts' minor 
```
image

### DIFF
--- a/.github/workflows/publishPackages.yml
+++ b/.github/workflows/publishPackages.yml
@@ -1,10 +1,10 @@
-name: Publish Single Package
+name: Publish Packages
 
 on:
   workflow_dispatch:
     inputs:
       package:
-        description: 'Package name'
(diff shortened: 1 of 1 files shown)


bachstatter (approved):
LGTM

PR #281, opened 2022-08-29
Selector & Pagination component

fritzschoff:
Bildschirmfoto 2022-08-29 um 16 58 01
Bildschirmfoto 2022-08-29 um 16 58 06
Bildschirmfoto 2022-08-29 um 16 58 07
Bildschirmfoto 2022-08-29 um 16 58 13


v3/theme/src/components/Pagination.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+  useEffect(() => {
+    if (currentPages[0] + pageSize >= maxLength) {
+      setCurrentPages([maxLength - pageSize, maxLength]);
+    } else {
+      setCurrentPages([currentPages[0], currentPages[0] + pageSize]);
+    }
+  }, [pageSize]);
+
+  useEffect(() => onChange(currentPages), [currentPages]);
```
bachstatter, line 45: I think it's bad passing an array to `useEffect`'s dependency array, I think it can behave quite strange. Is it possible to have this logic in event handlers instead?

bachstatter: Yea that's sounds right. Although I also now that it will complain the the number if items in the array changes.

v3/theme/package.json
```diff
@@ -25,6 +25,7 @@
     "lint:fix": "eslint --fix './src/**/*.{js,ts,tsx}'"
   },
   "dependencies": {
+    "@chakra-ui/icons": "^2.0.6",
```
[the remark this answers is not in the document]
fritzschoff: yeah was trying to do that but `deps:mismatched` complained about that. I should use version ..6. So I used that one

v3/theme/src/components/Pagination.tsx
fritzschoff: I mean I could do `.toString()`. Would that be okay for you guys?

v3/theme/package.json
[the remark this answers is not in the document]
fritzschoff: let me try again and come back to you if I got an error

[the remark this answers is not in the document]
fritzschoff: Updated to 2.0.9

v3/theme/src/components/Pagination.tsx
bachstatter: I haven't looked at the core logic deeply, but shouldn't it be possible to call the  `onChange`  callback directly in event handlers? 

You can do toString() but for eslint to be happy would need to do this:
```
const currentPagesString = currentPages.toString()
useEffect(() => onChange(currentPagesString.join()), [currentPagesString]);
```

fritzschoff: or just do in the deps array `currentPages.toString()`, no?

bachstatter: > or just do in the deps array currentPages.toString(), no?

If you just do that eslint will say 
`React Hook useEffect has a missing dependency: 'currentPages'. Either include it or remove the dependency array`

> Nah, better to have callbacks in smaller components and not use arrays in deps. Stringifying array means `array.join(',')`.

toString means `split(',')`.

bachstatter: > > or just do in the deps array currentPages.toString(), no?
> 
> If you just do that eslint will say `React Hook useEffect has a missing dependency: 'currentPages'. Either include it or remove the dependency array`
> 
> > Nah, better to have callbacks in smaller components and not use arrays in deps. Stringifying array means `array.join(',')`.
> 
> toString means `split(',')`.

haha join is right.. long day

fritzschoff: all good. Gotcha guys

fritzschoff: okay using now the string for the deps array. Also Jade just told me that if the page size changes, we should go back to start. Like you are two pages deep and you change to 16 items per page, you start again from the beginning with the new page size. Also returning now the zero based number to the consumer so he can directly use it in his slice method

0xjocke (commented):
👍

PR #284, opened 2022-08-30
Switch & Tooltip theming

fritzschoff:


v3/theme/src/components/Selector.tsx
```diff
@@ -15,12 +15,20 @@ export default function Selector({
...
   ...rest
 }: PropsWithChildren<SelectorProps>) {
+  let rightIconCloned: ReactElement;
+  let leftIconCloned: ReactElement;
+  if (rightIcon) {
+    rightIconCloned = cloneElement(rightIcon, { w: 3, h: 3 });
+  }
+  if (leftIcon) {
+    leftIconCloned = cloneElement(leftIconCloned, { w: 3, h: 3 });
+  }
```
fritzschoff, line 25: design says that we ALWAYS want to have this dimensions.

bachstatter: hmm yea its a little sad that we have to use cloneElement, but I guess not the end of the world

PR #288, opened 2022-09-01
Fix fetching propsals

0xjocke:
Council dilution isn't a thing anymore, so we can take the proposals straight from snapshot.

Using the dev build in staking:
image


noisekit (approved):
🤯

PR #292, opened 2022-09-02
Lint for unnecessary braces

0xjocke:


noisekit (approved):
Would be good for while monorepo

bachstatter:
> Would be good for while monorepo

Added to v3 too.
Since v3 and v2 are the only packages with `plugin:react/recommended` I only added it to v3

noisekit (approved):
🎉

PR #301, opened 2022-09-05
Build v2 libs

noisekit:
Created an example component as part of the UI with types/etc (not literally created but extracted an existing one). Because all of them depend on styles/common and few other things - extracted them too.

Updated webpack build to make everything work and checked that it all renders properly

So now we have a sample component building out of v2/components too. + few extra supporting libraries in v2/lib

PR is coming pretty large (but simply because of mass-replace of `styles/common` --> `@snx-v2/styles` and other libs).

Build is OK, Modal is working fine too.

Screen Shot 2022-09-05 at 14 39 35

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -58,7 +58,7 @@ module.exports = {
 
       settings: {
         react: {
-          version: '18.0.0',
+          version: '18.2.0',
         },
       },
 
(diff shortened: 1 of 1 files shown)


bachstatter (approved):
Looks good, a little hard to review but I checked it out and things seems to be working. Found [one issue](https://github.com/Synthetixio/js-monorepo/pull/306) unrelated to your PR,

PR #311, opened 2022-09-07
Transaction modal

0xjocke:
image

### DIFF
--- a/v2/components/TransactionModal/TransactionModal.tsx
+++ b/v2/components/TransactionModal/TransactionModal.tsx
@@ -1,0 +1,40 @@
+import React, { PropsWithChildren, ReactNode } from 'react';
+import {
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalBody,
+  ModalCloseButton,
+  Box,
+} from '@chakra-ui/react';
+import { theme } from '@synthetixio/v3-theme';
+import { useTranslation } from 'react-i18next';
+
+type Props = { isOpen: boolean; onClose: () => void; title: string; icon: ReactNode };
+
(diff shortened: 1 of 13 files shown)


v2/components/TransactionModal/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@snx-v2/TransactionModal",
+  "private": true,
+  "main": "index.tsx",
+  "version": "0.0.0",
+  "dependencies": {
+    "@chakra-ui/react": "^2.2.8",
+    "@snx-v2/icons": "workspace:*",
+    "@storybook/client-api": "latest",
```
bachstatter, line 9: @noisekit  This is a new dependency added to make dynamic stories better/easier to work with. Do you think I should I be adding that to the `v2/ui`s main package.json too?

bachstatter: `yarn deps:mismatched` just told me I should hehe..

v2/components/TransactionModal/TransactionModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+  ModalContent,
+  ModalHeader,
+  ModalBody,
+  ModalCloseButton,
+  Box,
+} from '@chakra-ui/react';
+import { theme } from '@synthetixio/v3-theme';
+import { useTranslation } from 'react-i18next';
+
+type Props = { isOpen: boolean; onClose: () => void; title: string; icon: ReactNode };
```
jmzwar, line 14: Do you prefer to use type or interface here? I don't mind but would be good for consistency.

jmzwar (approved):

PR #312, opened 2022-09-07
Feature/mint

jmzwar:
Screen Shot 2022-09-07 at 1 06 48 pm

Note - something up with storybook, perhaps due to unsupported props.


noisekit (approved):

v2/lib/formatters/number.ts
```diff
@@ -0,0 +1,5 @@
+export function numberWithCommas(value: string) {
```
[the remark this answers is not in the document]
bachstatter: Nope

bachstatter, line 1: I'd like to add a test for this. @noisekit  Do you have any thoughts for how jest should be done in this folder? Do I just add jest as a devDep?

bachstatter: Jest only exists in `/v2/ui/` so it wont pick up test files from `/v2/lib`.

v2/components/Mint/Mint.tsx
```diff
@@ -0,0 +1,207 @@
...
+}
+
+export const Mint = ({
+  snxBalance = wei(0),
+  susdBalance = wei(0),
+  gasPrice = wei(0),
+  exchangeRate = 0.25,
+}: MintProps) => {
+  const { t } = useTranslation();
+  const [val, setVal] = useState<string>('');
```
bachstatter, line 23: Should this just be props instead?

bachstatter: hmm no not neccesarily

jmzwar: The useState you mean? Possibly but hard to do the state management until we are integrating it. Happy to hoist stuff then.

bachstatter: > doesn't it infer type from default value? that `<string>` seems unnecessary


Correct `useState<string>('');` -> useState('');`

jmzwar: Did you change this @bachstatter? It's better to keep because you can only setState to a string. If you remove <string> you can set to whatever.

bachstatter: Nope typescript will still complain is you try setting it to something other than a `string`. It infers the type from initial state
image

0xjocke (approved):

bachstatter:
I gave up on trying to get the storybook ui work with Wei. :(

v2/components/Mint/Mint.tsx
jmzwar: Ok good to know

PR #313, opened 2022-09-08
Balance box

0xjocke:
Screen Shot 2022-09-08 at 11 03 20 am

### DIFF
--- a/v2/lib/formatters/number.test.ts
+++ b/v2/lib/formatters/number.test.ts
@@ -1,4 +1,4 @@
-import { numberWithCommas } from './number';
+import { formatNumber, formatNumberToUsd, numberWithCommas } from './number';
 
 describe('number', () => {
   test('numbersWithCommas', () => {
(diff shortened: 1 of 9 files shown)


noisekit (approved):

v2/lib/formatters/number.test.ts
```diff
@@ -13,4 +13,34 @@ describe('number', () => {
     expect(numberWithCommas('100000000')).toBe('100,000,000');
     expect(numberWithCommas('Not a number')).toBe('Not a number');
   });
+  test('formatNumber', () => {
+    expect(formatNumber(1)).toBe('1.00');
+    expect(formatNumber(10)).toBe('10.00');
+    expect(formatNumber(100)).toBe('100.00');
+    expect(formatNumber(1000)).toBe('1,000.00');
```
jmzwar, line 20: Nice

v2/lib/formatters/number.ts
```diff
@@ -3,3 +3,17 @@ export function numberWithCommas(value: string) {
...
+  return new Intl.NumberFormat('en-EN', optionsToUse).format(Number(val));
+};
+
+export const formatNumber = (value: number | string, options?: Intl.NumberFormatOptions) => {
+  const optionsWithDefault = {
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2,
+    ...options,
+  };
+  return new Intl.NumberFormat('en-EN', optionsWithDefault).format(Number(value));
```
jmzwar, line 18: Does this need to be able to take different format parameters? Thinking like Yen etc

jmzwar (approved):
Looks good. Just the one question on intl support.

v2/lib/formatters/number.ts
bachstatter: Yep sounds reasonable to me. Added!

jmzwar: That optionsWithDefault arg will handle the changes right? Pretty sure it is like for options: { style: 'currency', currency: 'JPY' }

jmzwar (approved):

v2/lib/formatters/number.ts
bachstatter: Ah yea you can change that however you want since we spread the options. The locales changes how commas are placed based on the country

PR #321, opened 2022-09-09
NO NEXT

noisekit:
Testing out deployments and builds in vercel.
~Only dashboard and mint and burn are currently added to the router. Rest will follow shortly~

All pages added to the router plus all the nested tabs checked

## Random deep nested route pages:

image

image

image

image

### DIFF
--- a/v2/ui/.storybook/main.js
+++ b/v2/ui/.storybook/main.js
@@ -1,6 +1,18 @@
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
 const isProd = process.env.NODE_ENV === 'production';
 
+// Make depcheck happy
+require.resolve('storybook');
+require.resolve('@chakra-ui/storybook-addon');
+require.resolve('@mdx-js/react');
+require.resolve('@storybook/addon-actions');
+require.resolve('@storybook/addon-docs');
+require.resolve('@storybook/addon-essentials');
+require.resolve('@storybook/addon-interactions');
+require.resolve('@storybook/addon-links');
+require.resolve('@storybook/builder-webpack5');
+require.resolve('@storybook/manager-webpack5');
+
 module.exports = {
   stories: ['../../**/**/*.stories.tsx'],
 
--- a/v2/ui/babel.config.js
+++ b/v2/ui/babel.config.js
@@ -1,3 +1,6 @@
+require.resolve('core-js');
+require.resolve('@babel/runtime-corejs3/core-js/date/now');
+
 module.exports = {
   presets: [
     require.resolve('@babel/preset-typescript'),
@@ -32,7 +35,7 @@ module.exports = {
         },
       },
     ],
-    ['@babel/transform-runtime', { corejs: 3 }],
+    [require.resolve('@babel/plugin-transform-runtime'), { corejs: 3 }],
   ],
 
   env: {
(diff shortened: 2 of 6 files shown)


v2/ui/sections/escrow/index.tsx
```diff
@@ -1,22 +1,25 @@
...
 
 const Index: React.FC = () => {
-  const router = useRouter();
-  const defaultTab = (router.query.action && router.query.action[0]) || EscrowPanelType.REWARDS;
+  const params = useParams();
+  const action = params.action || '';
+  // @ts-ignore
+  const currentTab = Object.values(EscrowPanelType).includes(action)
+    ? action
+    : EscrowPanelType.REWARDS;
```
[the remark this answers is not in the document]
bachstatter: Yea I think you could do `action as EscrowPanelType` but not sure it's better.. in general I dont like enums...

v2/ui/typings/missing-types.d.ts
```diff
@@ -4,3 +4,5 @@ declare module '@ensdomains/ensjs';
 declare module '@eth-optimism/watcher';
 declare module 'remarkable/linkify';
 declare module 'remarkable-external-link';
+declare module '@synthetixio/safe-import';
+declare module '@synthetixio/safe-import/safeImport';
```
bachstatter, line 8: Are there reasons why we want to import `safe-import/safeImport`? rather than `@synthetixio/safe-import`.. Also it should be fairly easy to add types to safeImport, since the native `import` have proper types

bachstatter (approved):
Left a few comments. I tried to be thorough and I think it all make sense to me.. A little sad to loose the speed of swc

v2/ui/bootstrap.tsx
```diff
@@ -0,0 +1,20 @@
...
+import 'tippy.js/dist/tippy.css';
+import './app.css';
+
+import './i18n';
+
+import App from './App';
+
+export async function bootstrap() {
+  const app = document.querySelector('#app');
+  const root = ReactDOM.createRoot(app as Element);
```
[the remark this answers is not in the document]
bachstatter: doesn't really matter. But this would give the dev a clearer error message (maybe) and remove the need for `as`
```
  const app = document.querySelector('#app');
  if (!app) throw Error('Missing div with #app id, check index.html');
  const root = ReactDOM.createRoot(app);
```

v2/ui/index.html
```diff
@@ -0,0 +1,139 @@
...
+      rel="preload"
+      as="font"
+      type="font/woff2"
+      crossorigin="anonymous"
+      href="/fonts/GT-America-Condensed-Bold.woff2"
+    />
+  </head>
+  <body>
+    <div id="app"></div>
+    <div id="global-loader">
```
jmzwar, line 67: Do we still need this global loader? How big are the chunks?

v2/ui/Routes.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+          <Route path="/synths" element={<SynthsPage />} />
+
+          <Route path="/gov" element={<GovPage />}>
+            <Route path=":spaceKey/:panel" element={<GovPage />} />
+          </Route>
+
+          <Route path="/earn" element={<EarnPage />}>
+            <Route path=":pool" element={<EarnPage />}>
+              <Route path=":action" element={<EarnPage />} />
```
jmzwar, line 60: May be an updated version but can't we use a catch all for path=/earn/:pool/:action?

bachstatter:
Storybook has some errors too

jmzwar (approved):

PR #323, opened 2022-09-12
Feature/burn

jmzwar:
Screen Shot 2022-09-12 at 10 47 57 am

Also added Alert styling
Screen Shot 2022-09-13 at 2 07 26 pm
Screen Shot 2022-09-13 at 2 27 24 pm
Screen Shot 2022-09-13 at 2 28 08 pm


v2/components/Burn/Burn.tsx
```diff
@@ -0,0 +1,255 @@
...
+              outline="none"
+              fontFamily="heading"
+              fontSize="xl"
+              fontWeight="black"
+              lineHeight="2xl"
+              color="white"
+              height="unset"
+              _focus={{ boxShadow: 'none !important' }}
+              _placeholder={{ color: 'whiteAlpha.700' }}
+            />
```
bachstatter, line 95: Have you given any thoughts on sharing a lot of think logic with mint? Not saying it should be the same component but maybe we should "dry" it up a little?

jmzwar: Yeah I think it makes sense to pull some of this into baseStyles

0xjocke (approved):
Nikta's suggestion should probably be implemented but otherwise it looks good. Might be worth doing another PR to DRY up mint and burn too

v2/ui/content/theme.ts
```diff
@@ -18,6 +84,20 @@ const Badge: ComponentStyleConfig = {
         cursor: 'pointer',
       },
     },
+    burn: {
+      width: `${100 / 3}%`,
```
bachstatter, line 88: width for burn 33% and for mint 100%?

noisekit (approved):

PR #333, opened 2022-09-14
Optimise build

noisekit:
- [x] Remove queries using snapshot from the context and instead use them directly on gov page
- [x] Ensure we include only one `bn.js` (it was like over 10 of them included before)


bachstatter (approved):
This is still a draft but it make sense to me!

PR #335, opened 2022-09-16
Minor bug fixes

0xjocke:
Before
Screen Shot 2022-09-16 at 11 54 18 am
After
Screen Shot 2022-09-16 at 11 53 59 am
Before
Screen Shot 2022-09-16 at 11 38 16 am
After
Screen Shot 2022-09-16 at 11 37 48 am

### DIFF
--- a/v2/ui/sections/loans/components/common.ts
+++ b/v2/ui/sections/loans/components/common.ts
@@ -12,8 +12,7 @@ export const FormContainer = styled(FlexDivColCentered)`
 `;
 
 export const InputsContainer = styled.div`
-  display: grid;
-  grid-template-columns: 1fr 1px 1fr;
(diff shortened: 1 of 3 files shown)


jmzwar:
Would it make sense to address that i18n bug in liquidations as well?

bachstatter:
> Would it make sense to address that i18n bug in liquidations as well?

Yea can you remind me how to replicate? Or fix it and push if you know where the problem is

jmzwar:
> > Would it make sense to address that i18n bug in liquidations as well?
> 
> Yea can you remind me how to replicate? Or fix it and push if you know where the problem is

Give me a sec and I'll fix

jmzwar (approved):

PR #336, opened 2022-09-16
Cypress

noisekit:
Base smoketest to load a homepage and check it was rendered

Currently requires build so takes some time
I will migrate from github actions and can use caching

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,7 +85,6 @@ jobs:
           background: true
       - run: yarn workspace @synthetixio/v2-ui run e2e:check
 
-
 workflows:
   version: 2.1
 


.circleci/config.yml
```diff
@@ -1,13 +1,89 @@
 version: 2.1
 
+orbs:
+  browser-tools: circleci/browser-tools@1.4.0
+
+commands:
+  yarn-install:
```
bachstatter, line 7: Maybe it's time to start using yarn 3 caching, is that waht plug and play is?

bachstatter:
Did you have a look at synpress? Would be convenient to have metamask installed

bachstatter (approved):

PR #340, opened 2022-09-16
Component testing

noisekit:
- [x] Move cypress into a separate package (to avoid clashing of webpack and babel configs and reduce installation for ui app)
- [x] Cypress-specific minimal webpack and babel configs
- [x] `CRatioHealthCard.cy.tsx` sample test
- [x] Enable component testing in CCI

Cypress clicks too fast for GIF, so here are just pinned screenshots.
Note:
- in the first test tooltips are tested
- in orange and red tests boundaries are changed and corresponding values are tested

image

image

image


drptbl:
Hey @noisekit, wouldn't it be easier to do it with `@synthetixio/synpress`? It should support component testing the same way cypress does and as an additional benefit - synpress would be already integrated with `js-monorepo` for e2e testing.

0xjocke (approved):

PR #341, opened 2022-09-18
Quickfix/pre release bugs

jmzwar:


.gitignore
```diff
@@ -35,7 +35,7 @@ pids
 
 /build/stats.json
 
-.yarn/*
+.yarn/**
```
bachstatter, line 38: I think we want this. We added the yarn cache on purpose to get no networks installs on CI

jmzwar: Ok - does add a lot of crud tho.

bachstatter: Yea. I think it's worth it. Good from a security standpoint and faster builds on CI. It's a documented feature of yarn v3

bachstatter (approved):

PR #350, opened 2022-09-20
remove useEffect

jmzwar:


bachstatter (approved):
Added a commit to navigate to home when a delegate wallet is picked (and delegate mode is activated)

PR #351, opened 2022-09-21
Inject custom Provider and Signer before app starts up

noisekit:
No actual test changes in this PR. Just a prep. But after opening home page it is possible to click around and get txns signed automatically

These ENVs are required before running:
```sh
export CYPRESS_TENDERLY_RPC_URL=""
export CYPRESS_WALLET_ADDRESS=""
export CYPRESS_WALLET_MNEMONIC=""
```

NOTE: for development it is convenient to work with single wallet on the same fork for easy debugging. To run in CI we will create fork automatically, generate and fund wallet, set ENVs, perform tests and kill the fork afterwards.

# getting a loan
auto-wallet in cypress

# claiming closed loan eth
auto-wallet in cypress - claim

# minting susd
auto-wallet in cypress - minting

### DIFF
--- a/v2/cypress/cypress/support/e2e.js
+++ b/v2/cypress/cypress/support/e2e.js
@@ -1,3 +1,5 @@
+import { ethers } from 'ethers';
+
 beforeEach(() => {
   cy.intercept('https://analytics.synthetix.io/matomo.js', { statusCode: 204 }).as('matomo');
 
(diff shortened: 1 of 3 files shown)


v2/cypress/cypress/support/e2e.js
```diff
@@ -13,7 +15,49 @@ beforeEach(() => {
...
+              return () => new Signer(win.ethereum);
+
+            case 'request':
+              return async ({ method, params }) => {
+                switch (method) {
+                  case 'eth_accounts':
+                  case 'eth_requestAccounts':
+                    return [Cypress.env('WALLET_ADDRESS')];
+                  case 'eth_chainId':
+                    return '0x1';
```
bachstatter, line 50: Doesn't need to happen now, but maybe this could come from the provider somehow, so we get automatic support for optimism based on the fork url

0xjocke (approved):
Proxy magic!! 🔥

PR #355, opened 2022-09-23
Feature/mint burn logic

jmzwar:
Few things to be discussed

=> Cache management - do we want to try and have larger hooks that we can refetch first from the cache? In this PR I have a useHeaderBalances which fetched sUSD and SNX data, but that data is also used in useDebtQuery.

=> Skeleton loading - at what level do we want to do it - see example in burn.

=> Need to check in on some of the calculations for mint/burn.


noisekit (approved):
All good but:
- no lodash please
- compose queries rather than contract calls

v2/lib/currency/currency.ts
```diff
@@ -0,0 +1,97 @@
+export type CurrencyKey = string;
+
+export enum Synths {
```
bachstatter, line 3: I dont think we should use Enums. They will get out of date. I think we should just keep synths and `Record<string, Wei | undefined>` Then it's up to the consumer to handle undefined values..  Or if we have good reasoning we should be using something like `loadSynthsByNameFromNetwork` https://github.com/Synthetixio/js-monorepo/blob/master/packages/queries/src/utils.ts#L31

v2/lib/useHeaderBalance/useHeaderBalance.ts
```diff
@@ -0,0 +1,73 @@
...
+export const processBalances = (balances: [SynthBalancesTuple, BigNumber]) => {
+  const [currencyKeys, synthsBalances, synthsUSDBalances] = balances[0];
+  const balancesMap: SynthBalancesMap = {};
+  let totalUSDBalance = wei(0);
+
+  currencyKeys.forEach((currencyKeyBytes32, idx) => {
+    const balance = wei(synthsBalances[idx]);
+
+    if (balance.gt(0)) {
+      const synthName = parseBytes32String(currencyKeyBytes32) as CurrencyKey;
```
bachstatter, line 29: If we agrees to just use string here we can remove the `as`

v2/lib/useHeaderBalance/useHeaderBalance.ts
```diff
@@ -0,0 +1,73 @@
...
+  const { data: Synthetix } = useSynthetix();
+
+  return useQuery(
+    ['walletBalances', 'header', networkId],
+    async () => {
+      if (!SynthUtil || !Synthetix || !walletAddress) {
+        throw Error('Query should not be enabled if contracts are missing');
+      }
+
+      return Promise.all([
```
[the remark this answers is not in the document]
jmzwar: Yeah that was the convo we were having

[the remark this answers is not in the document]
bachstatter: I think these two are ok to always call together, but we probably want use to just be calles `balances` and use it in `useDebtData` too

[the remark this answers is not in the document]
bachstatter: hmm we already have useSynthBalance that does the `SynthUtil.synthsBalances(walletAddress)` call? 
So I reckon add `useBalances` here and get the values with a `Object.values`?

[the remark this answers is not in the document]
jmzwar: Yeah I think that is probably the go - let me have another shot at it

v2/lib/currency/currency.ts
jmzwar: Shouldn't these always be updated though? We should be on top of what synths are available.

bachstatter: If you think it's worth it we should grab them from `@synthetixio/contracts` like `loadSynthsByNameFromNetwork` does, rather than hardcoding. The `@synthetixio/contracts` will have all the correct synths for each network. This enum is not correct since different network have different synths.

v2/components/Burn/Burn.tsx
```diff
@@ -38,8 +45,25 @@ export const Burn = ({
...
-  // TODO: Calculate c-ratio/debt clearing calcs
   const onBadgePress = (amount: ActiveBadge) => {
+    switch (amount) {
+      case ActiveBadge.max:
+        setVal(susdBalance.toString(2));
+        break;
+
+      case ActiveBadge.cRatio:
+        // TODO Figure out the logic for this
+        console.log('Burn c', targetCRatio, currentCRatio);
```
bachstatter, line 56: This it to burn back to to target c-ratio? 
`debtData.debtBalance.sub(debtData.issuableSynths)`

v2/components/Navigation/Navigation.tsx
```diff
@@ -247,3 +255,28 @@ export const Navigation = ({
     </Flex>
   );
 };
+
+export const NavigationUI = ({
```
bachstatter, line 259: This should be `Navigation` and the other one `NavigationUI`, since this it doing the data fetching and the other one render the ui

v2/components/Burn/Burn.tsx
jmzwar: Nice - I might put that in next pr

v2/lib/useSynthsBalances/useSynthsBalances.ts
```diff
@@ -0,0 +1,74 @@
...
+    totalUSDBalance,
+  };
+};
+
+export const useSynthsBalances = () => {
+  const { networkId, walletAddress } = useContext(ContractContext);
+  const { data: SynthUtil } = useSynthUtil();
+
+  return useQuery(
+    ['walletBalances', 'synths', networkId],
```
bachstatter, line 58: Add `walletAddress` here too, so that query get invalidated if user switches account in metamask

v2/ui/content/V2Burn.tsx
```diff
@@ -42,29 +66,39 @@ const V2Burn = () => {
...
+              currentCRatioPercentage={currentCRatio || 0}
+              targetCratioPercentage={targetCRatio || 0}
             />
-          </Box>
-          <Flex
+          </Skeleton>
+          <Skeleton
+            startColor="gray.900"
+            endColor="gray.700"
+            isLoaded={!isLoading}
```
bachstatter, line 93: This is pretty neat!

v2/ui/v2-components/Header/Header.tsx
```diff
@@ -1,6 +1,6 @@
 import { FC, useEffect, useState } from 'react';
 import { NetworkIdByName, NetworkId } from '@synthetixio/contracts-interface';
-import { Navigation } from '@snx-v2/Navigation';
+import { NavigationUI } from '@snx-v2/Navigation';
```
bachstatter, line 3: If you agree with my other comment, this can stay as Navigation. That's the pattern I've used else where. The UI version of the component is only used in tests and storybook. And the app uses the "normal" name

v2/components/Navigation/Navigation.tsx
jmzwar: Ah I see - understood

0xjocke (approved):

PR #358, opened 2022-09-26
Unflag options

0xjocke:
Screen Shot 2022-09-26 at 4 38 24 pm
Screen Shot 2022-09-26 at 4 21 17 pm
Screen Shot 2022-09-26 at 4 21 03 pm


noisekit (approved):

jmzwar (approved):
Looks good - UnflagOptions may need a wrapper for bg color but we can figure that out later.

bachstatter:
> Looks good - UnflagOptions may need a wrapper for bg color but we can figure that out later.

Hmm yea maybe. The design for that whole page has a different background (no pattern). Will figure out once it get integrated as a page

PR #360, opened 2022-09-27
Skeletons

0xjocke:
- Skeletons for CratioHealthCard, MainActionCards and balance box.
- A small fix to useSynthBalance
- Remove example mutation from main action cards


v2/components/BalanceBox/BalanceBox.tsx
```diff
@@ -31,30 +36,54 @@ export const BalanceBoxUi = ({
...
+        {snxBalance !== undefined ? (
+          <Text fontFamily="mono" fontWeight="extrabold" fontSize="sm" lineHeight="5">
+            {formatNumber(snxBalance)}
+          </Text>
+        ) : (
+          <Skeleton my={1} width={8} height={4} />
+        )}
+
+        {snxBalance !== undefined && snxPrice !== undefined ? (
+          <Text lineHeight="4" fontSize="xs" color="gray.500">
```
jmzwar, line 48: Hm, I think you can do Skeleton with a text child here rather than have conditional logic

bachstatter: I did try that, is was behaving a bit weird.  I think it was due to `formatNumber` needs a number so I had to `snxBalance && snxPrice && formatNumber(snxBalance * snxPrice)` which mean that no text was rendered and the skeletion would have 0 height.

jmzwar: Yeah - I think I had the same prob - what I did was just pass 0 if it was undefined - because skeleton has the isLoading prop it will never be seen

noisekit (approved):
ok tho

v2/components/MainActionCards/MainActionCards.tsx
```diff
@@ -165,7 +174,18 @@ const MaintainActionCard: React.FC<UiProps> = ({
...
+    Pick<
+      UiProps,
+      | 'liquidationCratioPercentage'
+      | 'targetCratioPercentage'
+      | 'currentCRatioPercentage'
+      | 'nextEpochStartDate'
+      | 'hasClaimed'
+    >
+  > & { snxPrice?: string }
+> = ({
```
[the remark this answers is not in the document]
bachstatter: hmm yea maybe I got a bit too fancy here

v2/components/BalanceBox/BalanceBox.tsx
```diff
@@ -64,14 +93,13 @@ export const BalanceBoxUi = ({
...
   const { data: exchangeRateData } = useExchangeRatesData();
-  if (!debtData || !exchangeRateData) return <p>Skeleton</p>;
 
   return (
     <BalanceBoxUi
-      snxPrice={Number(exchangeRateData.SNX?.toString())}
-      snxBalance={debtData.balance.toNumber()}
-      stakedSnx={debtData.collateral.toNumber()}
-      transferable={debtData.transferable.toNumber()}
+      snxPrice={exchangeRateData?.SNX?.toNumber()}
```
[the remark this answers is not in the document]
bachstatter: Yea I think its a case by case basis. When you have some text that you want to render it make sense for the UI component to have the logic.

It's also more covered by tests if the logic is in the UI component

v2/components/CRatioHealthCard/CRatioHealthCard.cy.tsx
```diff
@@ -2,6 +2,22 @@ import { Box } from '@chakra-ui/react';
...
+        <CRatioHealthCardUi
+          {...{
+            targetCratioPercentage: undefined,
+            liquidationCratioPercentage: undefined,
+            currentCRatioPercentage: undefined,
+          }}
+        />
+      </Box>
+    );
+    cy.contains('h2', 'Collateralization Ratio Health').should('be.visible');
```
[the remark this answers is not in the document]
bachstatter: yea that make sense to me. Will do in the future, and change these tests when/ if the copy changes

v2/components/MainActionCards/MainActionCards.tsx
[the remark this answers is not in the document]
bachstatter: Simplified it

v2/components/BalanceBox/BalanceBox.tsx
bachstatter: hmm yea could do that. But then I need to have the isLoaded logic in two places so I thought this was better

jmzwar, line 48: It'll work but will be good to align on a pattern for this in the future.

jmzwar (approved):

PR #362, opened 2022-09-27
Tenderly

noisekit:
Tool to create forks, get eth/snx and save/load fork checkpoints on Tenderly

See Readme for details/usage

Added usage to CI, creating a fork, filling necessary ENVs before e2e kicks in and then removing fork (regardless of steps successs)

image


0xjocke (approved):
Looks good, nice work! The only hting I was thinking about would be to add optimism support. Looks like not much have to change. Just have a networkId env var and import the optimism contract too `@synthetixio/contracts/build/mainnet/deployment/ProxyERC20.js`

PR #363, opened 2022-09-28
More CircleCI

noisekit:
- Move all tests and checks over to CCI
- Split out steps to reuse caches and parallelise better
- Remove `main` github actions workflow
- Add pubslishDev on-demand github actions workflow to publish all packages as `dev` (manual, only if we need it)
- Use cypress' own image that lets us to avoid the need for browsers/cypress installations

Parallelisation and caching

image


0xjocke (approved):
Nice, looks all good to me!!
I did think of an unrelated feature request. Ideally we should have a check on what branch the publish packages workflow is being run on. And we should only allow "proper semver" versions if the workflow is triggered with the master branch. (Doesn't have to go into this PR)

PR #368, opened 2022-09-29
More dynamic get SNX

noisekit:
Using Rewards distribution is not really viable as it turns out, so we are back to transferring funds directly from Rewards owner wallet.
Just made it more dynamic and avoid unnecessary ENV var

image

### DIFF
--- a/tools/tenderly/lib/getsnx.mjs
+++ b/tools/tenderly/lib/getsnx.mjs
@@ -1,18 +1,17 @@
 import { ethers, providers, utils } from 'ethers';
 import * as RewardsDistribution from '@synthetixio/contracts/build/mainnet/deployment/RewardsDistribution.js';
 import * as ProxyERC20 from '@synthetixio/contracts/build/mainnet/deployment/ProxyERC20.js';
-import * as Synthetix from '@synthetixio/contracts/build/mainnet/deployment/Synthetix.js';
 
 export async function getsnx(envs) {
-  const { TENDERLY_FORK_ID, TENDERLY_WALLET_ADDRESS } = envs;
-  if (!TENDERLY_FORK_ID) {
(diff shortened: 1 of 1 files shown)


tools/tenderly/lib/getsnx.mjs
```diff
@@ -22,15 +25,15 @@ export async function getsnx(envs) {
   // converts 1 ether into wei, stripping the leading zeros
   const tokenAmount = utils.hexValue(utils.parseEther('100').toHexString());
 
-  const unsignedTx = await contract.populateTransaction.transfer(
+  const unsignedTx = await contract.populateTransaction.addRewardDistribution(
     await signer.getAddress(),
```
bachstatter, line 29: Did you check that you actually got the SNX? I think it's a little more complicated. This would only add your address to receive rewards when rewards are distributed. I think we need another call to actually distribute the rewards. The `distributeRewards` method looks like a good candidate

jmzwar (approved):

tools/tenderly/random/cleanup-pending-distibutions.mjs
```diff
@@ -0,0 +1,59 @@
...
+    await previous;
+    const distribution = await RewardsDistributionContract.distributions(0);
+    console.log({
+      distributionToRemove: {
+        destination: distribution.destination,
+        amount: utils.formatUnits(distribution.amount),
+      },
+    });
+    const tx = await RewardsDistributionContractConnected.removeRewardDistribution(0);
+    const receipt = await tx.wait();
```
bachstatter, line 42: Haha I think this script is a little dangerous. If you remove all distributions the synthetix system will be a bit broken on the fork.  But I guess it is in a random folder so all good

bachstatter (approved):

PR #370, opened 2022-09-29
Tenderly CLI: removeMinimumStakeTime

noisekit:
image


jmzwar (approved):
I could actually use this rn

bachstatter (commented):
I would have liked a `setMinimumStakeTime`. But I guess you want to avoid args

PR #380, opened 2022-10-04
Generate types for Subgraph

noisekit:
To get typed hook generated we need to add a `x.query` file and run `yarn gql:gen` in `v2/ui`
I am not sure how it is going to work with v2/components though. Suspect we'll need to reiterate on the config

PS: make sure you exclude `.zip` packages from the PR review

image


v2/ui/queries/issueds/issueds.graphql
```diff
@@ -0,0 +1,7 @@
+query issueds($account: Bytes) {
```
bachstatter, line 1: If we need to write out each query I'm not use how much benefits this adds?  I could just write this add add it to the body of a fetch with useQuery?

v2/ui/queries/issueds/issueds.tsx
```diff
@@ -0,0 +1,51 @@
+// !!! DO NOT EDIT !!! Automatically generated file
+
+import * as Types from '../../schema.graphql';
+
+import { useQuery, UseQueryOptions } from '@tanstack/react-query';
+
+function fetcher<TData, TVariables>(query: string, variables?: TVariables) {
+  return async (): Promise<TData> => {
+    const res = await fetch('https://api.thegraph.com/subgraphs/name/synthetixio-team/synthetix', {
```
bachstatter, line 9: We do have a seperate subgraph for optimism https://api.thegraph.com/subgraphs/name/synthetixio-team/optimism-main/graphql. Any thoughts of how we work around that?

bachstatter: hmm yep I think that make sense. Wanna do an example for this one?

bachstatter: But this one is generated in a way that means we cant use it? Or we need to generate different ones for mainent /optimism.  I thought we said we shouldn't generate and instead write the query and import the types? So yea we do need a way to import the types from v2/lib.

bachstatter (approved):

PR #386, opened 2022-10-05
E2E test to borrow sUSD with ETH collateral

noisekit:
Few extra changes and fixes:
- Switch to use PK instead of MNEMONIC in Cypress (tenderly does not give us mnemonics and mnemonic cannot be derived from PK)
- Print CYPRESS_* variables when using `tenderly-fork` CLI
- Don't restart tests on file changes automatically, that's heavy

Screen Shot 2022-10-05 at 13 45 40

### DIFF
--- a/v2/cypress/cypress/e2e/loans.e2e.js
+++ b/v2/cypress/cypress/e2e/loans.e2e.js
@@ -5,6 +5,8 @@ describe('loans', () => {
     // Data fetching from tenderly can take some extra time
     const TIMEOUT_DATA_FETCH = 30000;
 
+    cy.viewport(1000, 1000);
+
     cy.visit('http://localhost:3000/loans');
     cy.get('[id="Borrow Synths-tab"]').should('be.visible');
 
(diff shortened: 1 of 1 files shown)


v2/cypress/cypress/e2e/loans.e2e.js
```diff
@@ -0,0 +1,29 @@
+describe('loans', () => {
+  it('Can borrow SNX with ETH collateral', () => {
```
bachstatter, line 2:
```suggestion
  it('Can borrow sUSD with ETH collateral', () => {
```

v2/ui/server.js
```diff
@@ -0,0 +1,52 @@
...
+      const ext = `${pathname.split('.').pop()}`.toLowerCase();
+      cache[pathname] = { type: mimeTypes[ext], data };
+      return send(response, cache[pathname]);
+    }
+    cache[pathname] = { type: 'text/html', data: index };
+    return send(response, cache[pathname]);
+  });
+});
+
+httpServer.listen(3000, '0.0.0.0', () => console.log(`Server is running at http://127.0.0.1:3000`));
```
bachstatter, line 52: Wow no express, nice

bachstatter (approved):

v2/cypress/cypress/e2e/loans.e2e.js
```diff
@@ -0,0 +1,42 @@
+describe('loans', () => {
+  it('Can borrow sUSD with ETH collateral', () => {
+    // TX confirmation can take a while
+    const TIMEOUT_TX = 120000;
+    // Data fetching from tenderly can take some extra time
+    const TIMEOUT_DATA_FETCH = 30000;
+
+    cy.visit('http://localhost:3000/loans');
+    cy.get('[id="Borrow Synths-tab"]').should('be.visible');
```
jmzwar, line 9: I don't think we've been adding data id's in new staking but we should probably prioritize.

jmzwar (approved):

PR #394, opened 2022-10-10
Move txnreducer to package

bachstatter:
This will be reused in useBurnMutation too

### DIFF
--- a/v2/lib/txnReducer/txnReducer.ts
+++ b/v2/lib/txnReducer/txnReducer.ts
@@ -1,0 +1,62 @@
+export type TransactionStatus = 'unsent' | 'prompting' | 'pending' | 'success' | 'error';
+
+type MintMutationState = {
+  error: Error | null;
+  modalOpen: boolean;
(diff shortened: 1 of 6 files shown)


jmzwar (approved):
yeah good idea

v2/lib/txnReducer/txnReducer.ts
```diff
@@ -0,0 +1,62 @@
+export type TransactionStatus = 'unsent' | 'prompting' | 'pending' | 'success' | 'error';
+
+type MintMutationState = {
```
bachstatter, line 3:
```suggestion
type MutationState = {
```

v2/lib/txnReducer/txnReducer.ts
```diff
@@ -0,0 +1,62 @@
...
+  error: Error | null;
+  modalOpen: boolean;
+  txnStatus: TransactionStatus;
+};
+
+export const initialState: MintMutationState = {
```
bachstatter, line 9: Should rename this

PR #398, opened 2022-10-12
Minting V2 e2e test

noisekit:
- Test minting of sUSD (will add burning when we have it working in v2)
- Few minor fixes for the UI + extra testids added
- Return populated transaction from the gas estimator to avoid doubling populateTransaction call (can be out of sync between gas estimation and actual txn sending)

https://user-images.githubusercontent.com/28145325/195501228-7381ae7e-78ef-4914-8599-423abd78dfb3.mp4

### DIFF
--- a/v2/lib/useGasOptions/useGasOptions.test.ts
+++ b/v2/lib/useGasOptions/useGasOptions.test.ts
@@ -19,11 +19,17 @@ const gasPricesMainnetMockData = {
     baseFeePerGas: wei(10, GWEI_DECIMALS, true).toBN(),
   },
 };
+
 const gasPricesOptimismMockData = {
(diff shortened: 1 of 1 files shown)


v2/lib/useMintMutation/useMintMutation.ts
```diff
@@ -64,9 +64,8 @@ export function useMintMutation(mintArgs: MintArgs) {
 
       try {
         dispatch({ type: 'prompting' });
-        const populatedTxn = await populateTransaction();
         const txn = await Synthetix.signer.sendTransaction({
-          ...populatedTxn,
+          ...populatedTransaction,
```
bachstatter, line 68: I was thinking of doing this too.. But if the user change the amount the gas query hook might still be loading and the sign of an out of date transaction? Not sure if this is true

jmzwar (approved):

PR #403, opened 2022-10-13
Add transaction modal for burn

bachstatter:
This PR is based on #402 pls have a look at that first.
Screen Shot 2022-10-13 at 6 06 40 pm
Screen Shot 2022-10-13 at 6 06 29 pm
Screen Shot 2022-10-13 at 6 06 07 pm


jmzwar (approved):
It looks fine - but I think we need to move the Transaction modals out into their own components for both burn/mint.

PR #404, opened 2022-10-14
E2E Burn

noisekit:
Based off of #402 and #403 

- [x] BLOCKED: waiting for #402 to be merged first
- [x] BLOCKED: waiting for #403 to be merged first
- [x] Pre-minting for burn test is done directly in Cypress test.
- [x] In this PR: I want to add unlock in there too to not depend on special scripts
- [x] update minting include necessary contract operations
- [x] snapshot/restore for each test to avoid any clashes between tests

Instead of using global test state and prime for all possible cases, each test now primes the data as it needs. This is achieved by adding cypress `tasks` for blockchain operations.
At this point I've added:
- getSnx
- mintSusd
- removeMinimumStakeTime
- snapshotSave / snapshotLoad

Eventually will refactor `tenderly` utils for better code reuse. And these cypress tasks will probably be simplified.

https://user-images.githubusercontent.com/28145325/196091009-b1c56b69-0adc-4c61-a8fe-82ad646b67df.mp4


v2/cypress/cypress/tasks/getSnx.js
```diff
@@ -0,0 +1,50 @@
+async function getSnx(amount = 100) {
```
bachstatter, line 1: Didn't we have a tendelry task for this?

bachstatter (approved):

v2/cypress/cypress/tasks/getSnx.js
bachstatter: Cool. The getSNX task is a bit flakey so would be nice to move to tenderly

PR #408, opened 2022-10-18
Txn modals refactor

bachstatter:
Move txn modal components to their own component


jmzwar (approved):
Nice, definitely cleans things up a bit

PR #418, opened 2022-10-20
Fix network switcher

sunnyvempati:
- Removed `window.ethereum` event handler.
- When user has no query param, default to whatever the active chain the connected wallet is on.
- When user has a query param with a chain name that isn't supported, default to goerli (configurable)
- When user changes chain on wallet OR via the rainbowkit button on the page, url and local chain are synced appropriately.

These are in conjunction to all the rules that already exist.  I'm also sure there's some way to clean up the logic, it's a bit confusing but i added plenty of comments to make it obvious.

### DIFF
--- a/v3/ui/src/App.tsx
+++ b/v3/ui/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { Suspense } from 'react';
+import React, { Suspense, useEffect } from 'react';
 import { Spinner, useColorMode } from '@chakra-ui/react';
 import { Route, Routes } from 'react-router-dom';
 import { DefaultLayout } from './layouts/Default';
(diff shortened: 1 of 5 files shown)


v3/ui/src/App.tsx
```diff
@@ -18,9 +18,12 @@ import { NotFoundPage } from './pages/404';
...
   const { colorMode, toggleColorMode } = useColorMode();
-  if (colorMode == 'light') {
-    toggleColorMode();
-  }
+
+  useEffect(() => {
+    if (colorMode === 'dark') {
+      toggleColorMode();
+    }
+  });
```
Rickk137, line 26: shouldn't it have a deps array?

Rickk137 (changes requested):

Rickk137 (approved):

v3/ui/src/App.tsx
sunnyvempati: done.  👍🏽

PR #413, opened 2022-10-19
Test coverage

noisekit:
- Code coverage and test reports for:
  - Cypress component testing
  - Cypress e2e testing
  - Jest
- Download and combine all coverage reports into a single one

Combined report: https://output.circle-artifacts.com/output/job/6de32a4d-6caa-421a-9847-2992b00d8f5b/artifacts/0/coverage/index.html

image

To find it:
image

PR #424, opened 2022-10-21
Restructure and optimise CCI build

noisekit:
Extracted this part out of my #413 PR as it was possible. 
- Full build is no longer required
- When running builds webpack will build all the monorepo code. It does add just a bit to the build time (as the amount of extra code is negligible), but makes it unnecessary to run `yarn build` before starting local server. Also will let me update Vercel and reduce deployment time significantly
- Because now all steps are in parallel, CCI passes in about 5-6 min (with yarn install) and in 3-4 min with cached node_modules
- For jest and typecheck we do `--no-private` build still (only publicly publishable packages), it is much faster than a full build

### DIFF
--- a/v2/cypress/cypress/tasks/getSnx.js
+++ b/v2/cypress/cypress/tasks/getSnx.js
@@ -1,8 +1,8 @@
-async function getSnx(amount = 100) {
-  const { ethers } = require('ethers');
-  const RewardsDistribution = require('@synthetixio/contracts/build/mainnet/deployment/RewardsDistribution.js');
-  const ProxyERC20 = require('@synthetixio/contracts/build/mainnet/deployment/ProxyERC20.js');
+import { ethers } from 'ethers';
+import * as RewardsDistribution from '@synthetixio/contracts/src/mainnet/deployment/RewardsDistribution';
+import * as ProxyERC20 from '@synthetixio/contracts/src/mainnet/deployment/ProxyERC20';
 
+export async function getSnx(amount = 100) {
   const provider = new ethers.providers.JsonRpcProvider(process.env.CYPRESS_TENDERLY_RPC_URL);
   const wallet = process.env.CYPRESS_WALLET_ADDRESS;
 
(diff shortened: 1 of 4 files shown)


v2/ui/webpack.config.js
```diff
@@ -30,7 +31,9 @@ function optimiseContracts() {
     )
     .concat([
       new webpack.NormalModuleReplacementPlugin(
-        new RegExp('/synthetix/publish/deployed/(kovan|local)'),
+        isTest
+          ? new RegExp('/synthetix/publish/deployed/(kovan|local|goerli|goerli-ovm|mainnet-ovm)')
```
bachstatter, line 35: We might want to write some tests for ovm, but I guess we dont right now

v2/ui/webpack.config.js
```diff
@@ -164,6 +182,24 @@ module.exports = {
...
+      [
+        'contracts-interface',
+        'optimism-networks',
+        'providers',
+        'queries',
+        'transaction-notifier',
+        'wei',
+        'generate-subgraph-query',
+        'v3-theme',
+      ].map(
```
bachstatter, line 195: Maybe this should read folder name in packges?

v2/ui/webpack.config.js
```diff
@@ -208,6 +244,15 @@ module.exports = {
...
 
   resolve: {
+    alias: {
+      '@synthetixio/contracts/build': '@synthetixio/contracts/src',
+      '@synthetixio/contracts-interface/build': '@synthetixio/contracts-interface/src',
+      '@synthetixio/optimism-networks/build': '@synthetixio/optimism-networks/src',
+      '@synthetixio/providers/build': '@synthetixio/providers/src',
+      '@synthetixio/queries/build': '@synthetixio/queries/src',
+      '@synthetixio/transaction-notifier/build': '@synthetixio/transaction-notifier/src',
+      '@synthetixio/wei/build': '@synthetixio/wei/src',
```
bachstatter, line 254: Wouldn't this use this aliases for prod build also?

bachstatter (approved):
Some thoughts/ comments left but overall make sense

v2/ui/webpack.config.js
bachstatter: Hmm yea. Maybe a a little readme note is enough.

v2/ui/webpack.config.js
bachstatter: ah I see! I for some reason thought it was for dev. But it make sense

PR #429, opened 2022-10-21
Add infura to providers

sunnyvempati:


v3/ui/src/utils/constants.ts
```diff
@@ -83,3 +83,12 @@ export const poolsData: Record<string, { name: string }> = {
...
 };
+
+export const INFURA_KEY = '23087ce9f88c44d1b1c54fd7c07c65fb';
+
+export const ALCHEMY_KEY_MAPPING: Record<number, string> = {
+  1: 'Yq-4rzpsP9Dz8mvjHRQ8lUrmNLNocuQs',
+  5: '6btiU37vkhQ8DJkdV4qggEWXehpgw2aI',
+  10: '-ktIzLrAY_Sf7AAlM1aTuZ-PObMop5sU',
+  420: 'lc1tSn8B8fmURkjmZLeyT1ibNpgqlsnr',
+};
```
sunnyvempati, line 94: we can move these into env vars later.  not a huge deal since these are public anyway and we're whitelisted.

drptbl: and.. these keys will be exposed anyway for everyone visiting our dapp tho (in source code and network requests)! :)

PR #431, opened 2022-10-24
UI improvements

jmzwar:


v2/components/Navigation/Navigation.tsx
```diff
@@ -289,10 +292,15 @@ export const Navigation = ({
...
-  });
+  const size = useBreakpointValue(
+    {
+      base: 'mobile',
+      md: 'desktop',
+    },
+    { fallback: 'md' }
+  );
+
+  const thing = 'hello';
```
jmzwar, line 303: oops

0xjocke (commented):
Looks good to me. 
You could remove some of the custom styling for `Tooltip` since we add those styles in the theme file

0xjocke (approved):

jmzwar:
> Looks good to me. You could remove some of the custom styling for `Tooltip` since we add those styles in the theme file



> Looks good to me. You could remove some of the custom styling for `Tooltip` since we add those styles in the theme file

There is a bug with the Tooltip - need to add the bg prop directly otherwise the arrow part of it doesn't get coloured

PR #434, opened 2022-10-26
Self liquidation page

bachstatter:
Will wire up the button in next PR
image

### DIFF
--- a/v2/components/SelfLiquidation/SelfLiquidation.tsx
+++ b/v2/components/SelfLiquidation/SelfLiquidation.tsx
@@ -87,21 +87,19 @@ export const SelfLiquidationUi: FC<{
           <Flex>
             <SNXIcon mt="5px" mr={2} />
             <Box>
-              <Text fontSize="2xl" fontWeight={800}>
-                {selfLiquidationPenaltySNX !== undefined ? (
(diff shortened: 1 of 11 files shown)


v2/components/SelfLiquidation/SelfLiquidation.tsx
```diff
@@ -1,65 +1,126 @@
...
+          alignItems="center"
+          bg="black"
+          w="58%"
+          pt={3}
+          px={4}
+          borderRadius="base"
+          borderWidth="1px"
+          borderColor="gray.900"
+        >
+          <CRatioProgressBar
```
bachstatter, line 62: Hey @noisekit, if I wanted to write a cypress component test for this file,  when it renders `CRatioProgressBar` it will try to fetch + require a query client.  

Any thoughts on the best pattern. Shallow render, mocking or intercept?

bachstatter: Yea I mean the fetching is done in a container, this just that the SelfLiquidationUi uses two other components which have their own react-queries.. I little cumbersome to use Intercept but I give it a go

jmzwar:
Might make sense to add a storybook component for this?

PR #441, opened 2022-10-27
Use png for curve icon

bachstatter:
- 6kb vs 274kb
Screen Shot 2022-10-28 at 7 58 16 am


jmzwar:
Did you check if this looks decent on retina?

I am always super reluctant to use raster images.

bachstatter:
> Did you check if this looks decent on retina?
> 
> I am always super reluctant to use raster images.

I looks a little bit worse. I exported it 2x the size though

jmzwar:
Ok let me put my glasses on

jmzwar:
It does look quite rasterised. 
I get that the file is huge but there is no performance hit in rendering it - can we just leave it for now?

bachstatter:
> It does look quite rasterised. I get that the file is huge but there is no performance hit in rendering it - can we just leave it for now?

Hmm yea maybe she should just use it as a real SVG + img tag, so we dont add 270kb extra for react/js to parse

jmzwar:
I'd be happy with that. Leave it to me and I can add it here

bachstatter (commented):
Looks good to me

jmzwar:
> Isn't it still way too big?

Yeah, but it's all lines, so raster makes it look pixellated.

PR #442, opened 2022-10-28
Wallet modal

bachstatter:
Some differences compared to the design, need to have a chart with Jade.. But at least a start 
Screen Shot 2022-10-28 at 12 46 45 pm

Things to chat with Jade about:
- Icons, the design is using perps icons
- The description under the synth name, is currently not included. If we want we could use the description from here: https://github.com/Synthetixio/synthetix/blob/develop/publish/assets.json?
- "Connected with MetaMask" note, not sure how much value that adds and also not sure how easy it is to know what wallet is used
- Leaving out Delegated and Watch Wallet. I think we need designs for these. currently they opens in a modal. It would be a little weird to open a modal from a modal.
- How should this popup look when we're not connected? Should the Account menu item be hidden?
-  We will only show synths if you have any balance, and maximum 5 of them. SNX is always on top, even if you dont have any. Is this the expected behavoiur?

### DIFF
--- a/v2/components/Navigation/Navigation.tsx
+++ b/v2/components/Navigation/Navigation.tsx
@@ -1,3 +1,4 @@
+import { useState } from 'react';
 import {
   Button,
   Center,
@@ -9,7 +10,6 @@ import {
   Text,
   useBreakpointValue,
 } from '@chakra-ui/react';
-
 import { NetworkId, NetworkIdByName } from '@synthetixio/contracts-interface';
 import {
   ChevronDown,
@@ -38,6 +38,7 @@ import { useSynthsBalances } from '@snx-v2/useSynthsBalances';
 import { EpochPrice } from '@snx-v2/EpochPrice';
 import { useExchangeRatesData } from '@snx-v2/useExchangeRatesData';
 import { useFeePoolData } from '@snx-v2/useFeePoolData';
+import { WalletModal } from '@snx-v2/WalletModal';
 
 interface NavigationProps {
   currentNetwork: NetworkId;
@@ -48,6 +49,7 @@ interface NavigationProps {
   isLoading: boolean;
   snxBalance: Wei;
   sUSDBalance: Wei;
+  disconnectWallet: () => Promise<void>;
 }
 
 const activeIcon = (currentNetwork: NetworkId) => {
@@ -75,9 +77,10 @@ export const NavigationUI = ({
   isLoading,
   snxBalance,
   sUSDBalance,
+  disconnectWallet,
 }: NavigationProps) => {
   const { t } = useTranslation();
-
+  const [isWalletModalOpen, setIsWalletModalOpen] = useState(false);
   const { name, icon } = activeIcon(currentNetwork);
   const navigate = useNavigate();
 
@@ -257,7 +260,7 @@ export const NavigationUI = ({
                 <Text ml={2}>{t('common.wallet.menu.gov')}</Text>
               </Center>
             </MenuItem>
-            <MenuItem onClick={() => navigate('/')}>
+            <MenuItem onClick={() => setIsWalletModalOpen(true)}>
               <Center>
                 <WalletIcon color="white" />
                 <Text ml={2}>{t('common.wallet.menu.wallet')}</Text>
@@ -276,6 +279,11 @@ export const NavigationUI = ({
           </MenuList>
         </Menu>
       </Flex>
+      <WalletModal
+        isWalletModalOpen={isWalletModalOpen}
+        setIsWalletModalOpen={setIsWalletModalOpen}
+        disconnectWallet={disconnectWallet}
+      />
     </Flex>
   );
 };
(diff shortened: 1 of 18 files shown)


v2/components/WalletModal/WalletModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+  isWalletModalOpen: boolean;
+  setIsWalletModalOpen: (x: boolean) => void;
+  disconnectWallet: () => Promise<void>;
+  walletAddress: string | null;
+  networkId: number | null;
+  balances?: BalanceObject[];
+}> = ({
+  isWalletModalOpen,
+  setIsWalletModalOpen,
+  disconnectWallet,
```
jmzwar, line 45: Can we grab disconnect from global state here? Prevent the prop drilling.

v2/components/WalletModal/WalletModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+  balances,
+}) => {
+  const { t } = useTranslation();
+  const [copiedAddress, setCopiedAddress] = useState<boolean>(false);
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    let intervalId: number | undefined;
+    if (copiedAddress) {
+      intervalId = Number(
```
jmzwar, line 57: Do we need this? I thing chakra has a useClipboard

jmzwar:
Could we get a storybook component for this as well pls :)

v2/components/WalletModal/WalletModal.tsx
bachstatter: Ah can have a look at that. I just copy pasted from another place in the app

v2/components/WalletModal/WalletModal.tsx
bachstatter: hmm some prop drilling required since we can import the Connector in the components packages.

jmzwar: We can't import connector in the packages you mean? Is this that issue of multiple contexts?

I guess leave it for now but we will need a way to solve it.

v2/components/WalletModal/WalletModal.tsx
bachstatter: Worked great

v2/components/Navigation/Navigation.tsx
```diff
@@ -75,9 +77,10 @@ export const NavigationUI = ({
   isLoading,
   snxBalance,
   sUSDBalance,
+  disconnectWallet,
 }: NavigationProps) => {
   const { t } = useTranslation();
-
+  const [isWalletModalOpen, setIsWalletModalOpen] = useState(false);
```
jmzwar, line 83: I also forgot to mention there is also this
https://chakra-ui.com/docs/hooks/use-disclosure

v2/lib/useSynthsBalances/useSynthsBalances.ts
```diff
@@ -42,8 +45,8 @@ export const processSynthsBalances = (balances: SynthBalancesTuple) => {
   return {
     balancesMap: balancesMap,
     balances: orderBy(
-      Object.values(balancesMap).filter((val) => val !== null && val !== undefined),
-      (balance) => balance?.usdBalance.toNumber(),
+      Object.values(balancesMap).filter(notNill),
```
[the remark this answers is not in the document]
bachstatter: `notNil` is a type guard without it the type for the array incorrectly includes `undefined`.

With `Boolean`:
Screen Shot 2022-10-28 at 2 55 09 pm

With `notNil`
Screen Shot 2022-10-28 at 2 54 52 pm

v2/components/icons/AvatarIcon/AvatarIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import { Icon, IconProps } from '@chakra-ui/react';
```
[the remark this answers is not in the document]
bachstatter: I think we discussed this and decided not to use it?

v2/components/Navigation/Navigation.tsx
bachstatter: hmm yea could use that. But not too big of a difference?

jmzwar: Not really, has some toggle state, but this is fine.

I'll probably try to use it in the future just for standardisation reasons.

v2/components/WalletModal/WalletModal.tsx
bachstatter: Yea I think we need to move the connector to a package. But it was quite tricky to do if I remember correct

v2/components/Navigation/Navigation.tsx
bachstatter: Fair enough, using it now

jmzwar (approved):
nice

PR #443, opened 2022-10-28
Untenderly

noisekit:
- Add foundry and anvil to the CI
- Replace all the tenderly rpc calls with local anvil
- Fix few things in `Loans` 
  - Don't calc cratio if amount is 0 (avoid div by 0 contract error)
  - Invalidate synths queries after modifying loans (refetch synths balances)
  - Enable full `loans.e2e` test (open loan, close loan, redeem collateral)
- Add Cypress logs printing to the console so we can see what is going on while it runs

### DIFF
--- a/v2/cypress/cypress/lib/printBrowserLogs.js
+++ b/v2/cypress/cypress/lib/printBrowserLogs.js
@@ -78,6 +78,8 @@ export function printBrowserLogs(browser = {}, launchOptions) {
 
         cdp.on('disconnect', () => {
           console.log('Chrome Debugging Protocol disconnected');
+          console.log('Reconnecting...');
+          setTimeout(tryConnect, 100);
         });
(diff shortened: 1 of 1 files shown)


v2/cypress/cypress/support/e2e.js
```diff
@@ -30,9 +28,11 @@ beforeEach(() => {
     class Signer {
       constructor(provider) {
         this._isSigner = true;
-        this.getAddress = async () => Cypress.env('WALLET_ADDRESS');
+        this.getAddress = async () => '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266';
```
bachstatter, line 31: Maybe this should live in a module and be named something nice, like anvil wallet?

bachstatter (approved):

PR #452, opened 2022-10-31
UI/jade feedback

jmzwar:
Mainly changes to action cards but includes Settings


v2/components/MainActionCards/MainActionCards.tsx
```diff
@@ -201,83 +216,84 @@ const CollectActionCard: React.FC<{
...
     liquidationCratioPercentage,
     targetCratioPercentage,
     currentCRatioPercentage,
   });
-  const isStaking = currentCRatioPercentage > 0;
+
+  const isStaking = currentCRatioPercentage && currentCRatioPercentage > 0;
   const canClaim = !hasClaimed && variant === 'success';
+
+  console.log('is loading', isLoading);
```
bachstatter, line 230: missed log

v2/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,158 @@
+import {
```
bachstatter, line 1: This is currently not used? I think we can release the first version without this

v2/components/MainActionCards/MainActionCards.tsx
```diff
@@ -1,198 +1,213 @@
...
+  icon: ReactNode;
+  disabled: boolean;
+  Content: JSX.Element | null;
+  buttonVariant?: 'solid' | 'outline' | 'link';
+  buttonText: string;
+  buttonAction: () => void;
+  testId: string;
+}
+
+export const Card = ({
```
bachstatter, line 34: This is a nicer abstraction 👍

v2/components/Settings/Settings.tsx
jmzwar: Yeah - I added it but I think we can release without implementing it in ui

bachstatter (approved):
Remove some logs, fix CI and then good to go

v2/components/MainActionCards/MainActionCards.tsx
```diff
@@ -1,198 +1,213 @@
...
+      icon={<StakeIcon disabled={isLoading} />}
+      disabled={isLoading}
+      buttonText={
+        isStaking
+          ? t('staking-v2.main-action-cards.stake-link-button')
+          : t('staking-v2.main-action-cards.stake-main-button')
+      }
+      Content={null}
+      buttonVariant={isStaking ? 'link' : isLoading ? 'link' : 'solid'}
+      testId={isStaking ? 'main stake button' : 'start staking button'}
```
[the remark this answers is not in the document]
jmzwar: We can put it directly on the card - but it'll still be a ternary depending on isStaking

bachstatter:
> The abstraction of card feels a bit too early in my opinion and we could do with some copy paste for a while. Testid that is passed down through props is a clear indication to me that abstraction is not yet formed to be "abstract enough", this will be more painful to support. But this is obviously not a blocker for feature if time constrained

I think James was already feeling some pain with previous abstractions. It was hard to just add a disabled prop. ( the reason for why he did the card abstraction)

PR #453, opened 2022-11-01
Burn fixes

bachstatter:
The burn UI now accurately calculates unstaking amount taking your c-ratio into consideration.
If you burning to c-ratio unstaking will always be 0.

I also refactored the calculations a bit. They now just work with numbers. Input validation (NaN checks) is dont in the ui.
In general that shouldn't happen since the input field wont allow you to type non numbers.

Happy to talk through these changes if any of you wants to dive deeper.

Screen Shot 2022-11-02 at 10 04 24 am

### DIFF
--- a/v2/components/Burn/Burn.tsx
+++ b/v2/components/Burn/Burn.tsx
@@ -402,18 +402,23 @@ export const Burn: FC<{ delegateWalletAddress?: string }> = ({ delegateWalletAdd
     if (!debtData || !susdBalance || !snxPrice) return;
     switch (badgeType) {
       case 'toTarget':
-        const burnAmount = Wei.max(debtData.debtBalance.sub(debtData.issuableSynths), wei(0));
-        const burnAmountString = formatNumber(burnAmount.toNumber());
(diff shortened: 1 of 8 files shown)


v2/components/Burn/Burn.tsx
```diff
@@ -249,11 +251,13 @@ export const BurnUi = ({
             </Badge>
           </Flex>
         </Box>
-        {activeBadge === 'toTarget' && (
+        {snxUnstakingAmount === '0.00' && (
```
jmzwar, line 254: Will this condition always hit? What if I type 0

jmzwar (approved):
It pretty much makes sense but is getting a bit complicated. I think it's alright because we have the tests but would make sense in the future to try and simplify some of the logic.

Just the one question on '0.00'

v2/components/Burn/Burn.tsx
bachstatter: hmm yea if you type 0 it wont be hit.. I guess we want that

PR #457, opened 2022-11-02
Opt in button

bachstatter:
Screen Shot 2022-11-02 at 11 46 03 am

image


jmzwar (approved):
Looks fine - do we need to check on copy?

bachstatter:
> Looks fine - do we need to check on copy?

Hmm it's still hidden behind a query param. So I say we can deploy and let cav/jade play around with it

jmzwar:
> > Looks fine - do we need to check on copy?
> 
> Hmm it's still hidden behind a query param. So I say we can deploy and let cav/jade play around with it

Sounds good

PR #460, opened 2022-11-02
Add back weth-snx pools page

bachstatter:
Screen Shot 2022-11-02 at 4 00 05 pm


jmzwar (approved):
Hmm wonder how we lost this

bachstatter:
> Hmm wonder how we lost this

I guess there were no links to the page so we didn't add it when we moved to react-router?

PR #465, opened 2022-11-03
card changes

jmzwar:
Screen Shot 2022-11-03 at 10 11 21 pm

=> Add connect to first card
=> Better active state
=> Darker background colour loading/disbaled state
Screen Shot 2022-11-03 at 10 17 02 pm


bachstatter (approved):

bachstatter:
I think we can merge this. The different states are a lot better tested.

jmzwar:
> I think we can merge this. The different states are a lot better tested.

Yeah looking a lot better.

PR #466, opened 2022-11-04
Generate contracts

bachstatter:


bachstatter:
Merging this as e2e are failing when using the old contracts

PR #467, opened 2022-11-06
Target threshold

bachstatter:
The FeePool contract (where we claim) takes a threshold ( called `targetThreshold` on the `SystemSettings` contract) into account. If your c-ratio is below target but within the threshold you can claim. There's no threshold for liquidations.

We already fetched `targetThreshold` in the debtData hook. So this PR just make sure we take it into account in `getHealthVariant`

### DIFF
--- a/v2/components/Mint/MintHeader.tsx
+++ b/v2/components/Mint/MintHeader.tsx
@@ -69,6 +69,7 @@ export const MintHeaderUi: FC<{
   liquidationRatioPercentage?: number;
   targetCRatioPercentage?: number;
   currentCRatioPercentage?: number;
+  targetThreshold?: number;
   collateral?: number;
   debtBalance?: number;
   SNXRate?: number;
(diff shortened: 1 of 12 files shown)


v2/components/Mint/MintHeader.tsx
```diff
@@ -162,6 +164,7 @@ export const MintHeaderUi: FC<{
                   liquidationCratioPercentage={liquidationRatioPercentage || 0}
                   currentCRatioPercentage={currentCRatioPercentage || 0}
                   targetCratioPercentage={targetCRatioPercentage || 0}
+                  targetThreshold={targetThreshold || 0.01}
```
jmzwar, line 167: Hmm what is the reasoning behind this?

jmzwar: Otherwise all pretty clear

bachstatter: hmm yea I could just pass 0. It is currently set to 0.01 and unlikely to change.. But maybe it's still clearer to just pass 0 when loading

jmzwar (approved):

PR #477, opened 2022-11-08
We do not require a sUSD balance to calculate max | target

bachstatter:
Also a link update for the help button

### DIFF
--- a/v2/ui/content/V2Home.tsx
+++ b/v2/ui/content/V2Home.tsx
@@ -79,7 +79,7 @@ const V2Home = () => {
             </Box>
             <Box mt={4}>
               <BoxLink
-                href="https://synthetix.io/"
+                href="https://synthetix.io/guides"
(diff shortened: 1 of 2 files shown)


v2/ui/content/V2Home.tsx
```diff
@@ -79,7 +79,7 @@ const V2Home = () => {
             </Box>
             <Box mt={4}>
               <BoxLink
-                href="https://synthetix.io/"
+                href="https://synthetix.io/guides"
```
jmzwar, line 82: Should we move this to lib/Constants?

bachstatter: ah yep you already added that if I remember correctly

jmzwar: Yeh, EXTERNAL_LINKS.CMS.Home

jmzwar (approved):

PR #481, opened 2022-11-09
Maincard updates

bachstatter:
- Dont show burn badge when rewards are claimed
- Always make sure users who are staking can go to the claim page
- Update tests
- Add a story for each test case


jmzwar (approved):
nice. good to see the stories

PR #482, opened 2022-11-11
initial claim card

jmzwar:
Rewards state as also used in wallet.
Also a fix for utility card


bachstatter (approved):
Looks good. But would be sligthly nive if it didn't have `reward` in the name. I will use it for debt balance and other non rewards stats

PR #485, opened 2022-11-12
Wallet

bachstatter:
Wallet page. 

I think it's in a good enough state to merge.  But some things missing.

Screenshot 2022-11-12 at 5 58 22 pm
- Tab buttons should be styled slightly different
- Add the other asset table
- Sorting. Doesn't look like sorting is supported with chakras table.. Not sure if needed for now though

### DIFF
--- a/v2/components/StatBox/StatBox.tsx
+++ b/v2/components/StatBox/StatBox.tsx
@@ -1,24 +1,27 @@
-import { Flex, Text } from '@chakra-ui/react';
+import { Flex, FlexProps, Text } from '@chakra-ui/react';
 
-interface RewardsStatProps {
+export const StatBox = ({
+  label,
+  amount,
+  containerStyles,
+}: {
   label: string;
   amount: string;
-  align?: 'start' | 'center' | 'end';
-}
-
-export const StatBox = ({ label, amount, align = 'center' }: RewardsStatProps) => {
+  containerStyles?: FlexProps;
+}) => {
   return (
     <Flex
-      alignItems={align}
+      alignItems="center"
       flexDirection="column"
       w="33%"
       maxW="325px"
       py="14px"
       px={6}
-      borderRadius="xl"
+      borderRadius="base"
       borderWidth="1px"
       borderColor="gray.900"
       bg="whiteAlpha.50"
+      {...containerStyles}
     >
       <Text fontFamily="heading" fontWeight="semibold" fontSize="sm" color="gray.500">
         {label}
--- a/v2/components/Wallet/Wallet.tsx
+++ b/v2/components/Wallet/Wallet.tsx
@@ -1,0 +1,13 @@
+import { Heading } from '@chakra-ui/react';
+// import { useGetDSnxBalance } from '@snx-v2/useDSnxBalance';
+// import { useSynthsBalances } from '@snx-v2/useSynthsBalances';
+
+const WalletUi = () => {
+  return <Heading>Hello world</Heading>;
+};
+export const Wallet = () => {
+  // const { data: synthBalances } = useSynthsBalances();
+  // const { data: dSNXBalance } = useGetDSnxBalance();
+
+  return <WalletUi />;
+};
--- a/v2/components/Wallet/Wallet.tsx
+++ b/v2/components/Wallet/Wallet.tsx
@@ -1,6 +1,7 @@
 import { Box, Tabs, TabList, Tab, TabPanels, TabPanel, Heading } from '@chakra-ui/react';
 import { ArrowTopRight } from '@snx-v2/icons';
 import { useParams, useNavigate } from 'react-router-dom';
+import { WalletBalances } from '@snx-v2/WalletBalances';
 
 const walletPages = [
   'balances',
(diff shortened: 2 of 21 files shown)


v2/components/WalletBalances/WalletBalances.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { useExchangeRatesData } from '@snx-v2/useExchangeRatesData';
+import { SynthBalance, useSynthsBalances } from '@snx-v2/useSynthsBalances';
+import Wei from '@synthetixio/wei';
+import { formatPercent, formatNumberToUsd } from '@snx-v2/formatters';
+import { StatBox } from '@snx-v2/StatBox';
+import { useTranslation } from 'react-i18next';
+
+const WalletBalancesUi: React.FC<{
+  totalSynthBalance?: number;
+  dSNXBalance?: number;
```
jmzwar, line 29: Do you not like breaking this out into its own definition? I find it harder to read like this.

v2/components/StatBox/StatBox.tsx
```diff
@@ -1,24 +1,27 @@
...
+  amount,
+  containerStyles,
+}: {
   label: string;
   amount: string;
-  align?: 'start' | 'center' | 'end';
-}
-
-export const StatBox = ({ label, amount, align = 'center' }: RewardsStatProps) => {
+  containerStyles?: FlexProps;
```
jmzwar, line 10: Makes sense

v2/components/Wallet/Wallet.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <Heading size="md" textAlign="center">
+        {t('staking-v2.wallet.my-account')}
+      </Heading>
+      <Tabs
+        variant="soft-rounded"
+        colorScheme="cyan"
+        index={tabIndex}
+        onChange={(tabIndex) => {
+          const newTab = walletPages[tabIndex];
+          if (pagesStillUsingOldDesign.includes(newTab)) {
```
jmzwar, line 42: This feels hacky? Can we do it in a more declarative way?

v2/ui/content/theme.ts
```diff
@@ -311,6 +311,26 @@ const Menu: ComponentMultiStyleConfig = {
 const Divider: ComponentStyleConfig = {
   baseStyle: { borderColor: 'gray.900' },
 };
+const Table: ComponentMultiStyleConfig = {
```
jmzwar, line 314: Nice

jmzwar (commented):
Generally looks ok, I think I'd check with Jade first on Loading states before merging - it could change the structure a bit.

v2/ui/content/theme.ts
jmzwar: Can we do similar with Tabs here?

v2/components/Wallet/Wallet.tsx
bachstatter: yea I agree. Made it a bit more declarative

v2/components/WalletBalances/WalletBalances.tsx
bachstatter: hmm yea I guess I could. I'm quite use to reading it, and it one less thing to name and less scrolling up to see what's available. But definitely no strong opinions about it

v2/ui/content/theme.ts
bachstatter: Yep that was the plan. But they already defined custom styles for tabs [in v3](https://github.com/Synthetixio/js-monorepo/blob/master/v3/theme/src/theme.ts#L293)  .  So want to check with jade if I should change it for v3 too

bachstatter:
> Generally looks ok, I think I'd check with Jade first on Loading states before merging - it could change the structure a bit.

Yea can do some loading styled before merging

v2/ui/content/theme.ts
bachstatter: Tabs problem resolved

bachstatter:
Loading added too

v2/components/Wallet/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@snx-v2/Wallet",
```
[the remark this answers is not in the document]
bachstatter: Hmm yea I followed the pattern from Mint and Burn. Right not the components called Page only exists in the `ui` folder

v2/lib/SynthIcons/getPngSynthIconUrl.ts
```diff
@@ -0,0 +1,4 @@
+// TODO we should come up with a better solution for this...
+export const getPngSynthIconUrl = (currencyKey: string) => {
+  return `https://raw.githubusercontent.com/Synthetixio/synthetix-assets/master/synths/png/${currencyKey}.png`;
```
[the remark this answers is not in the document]
bachstatter: Yea the repo do have an npm package. You can import the icons like this:
`import sBTC from '@synthetixio/assets/synths/sBTC.svg';`
But I guess doing dynamic import doesn't really work with a dynamic string :(

PR #512, opened 2022-11-21
C ratio progress bar

bachstatter:


jmzwar (commented):
Looks good - seeing a zIndex issue with tooltips though. (See storybook)

bachstatter:
> Looks good - seeing a zIndex issue with tooltips though. (See storybook)

Nice catch, fixed

PR #513, opened 2022-11-21
Improve CSP

noisekit:
I've walked over all the v2 pages and most v1. Onboard wallet is still in question. There might be some requests going for some wallets which I haven't hit.

Clarification: in CI we run full build in prod mode and then check git diff. This ensures that changes to `csp.js` are matching `vercel.json`


bachstatter (approved):
Looks good. I guess the CI would just validate that there are no changes rather than actually mutate the vercel file.. But no biggie.. And as we talked about lets isolate this change in a seperate release

PR #518, opened 2022-11-22
Account navigation

0xjocke:
- If your on the home page and switch to an address with accounts -> navigate to account page. Previously it only worked on reload.
- If you're on a page that requires account and switch to a wallet without accounts -> navigate to home

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -69,7 +69,7 @@ module.exports = {
       rules: {
         quotes: 'off',
         'react/jsx-curly-brace-presence': ['error', { props: 'never', children: 'never' }],
-
+        'react/prop-types': 'off', // use ts
         // react-hooks
         'react-hooks/rules-of-hooks': 'error',
         'react-hooks/exhaustive-deps': 'error',
--- a/v3/ui/src/App.tsx
+++ b/v3/ui/src/App.tsx
@@ -16,7 +16,7 @@ import { Market } from './pages/markets/Market';
 import { Pool } from './pages/pools/pool';
 import { Teleporter } from './pages/teleporter/Teleporter';
 import { NotFoundPage } from './pages/404';
-import RequireAccount from './components/accounts/RequireAccount';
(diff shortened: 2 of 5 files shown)


noisekit (approved):

sunnyvempati:
@bachstatter it's actually by design that we want users to navigate to account pages without needing to be signed in.  the idea is anyone can go to anyone's account and view the data; we have some conditionally rendering components within the components that remove functionality if you don't have an account attached (there's more work to be done regarding this on the page itself; but think of this is as a view vs. write page)  cc @noahlitvin

noahlitvin:
👍 This is correct. Views for all entities in all networks should be accessible regardless of wallet status. We just have a little bit of logic to decide where to kick you when loading the root path.

bachstatter:
Yea ok we were all a little confused when switching wallets.  
So to clarify:
Say I have two wallets with staking positions. If I'm connected to one of them and switch to the other wallet the app will stay on the same page, (and I guess be in read only) for the previous wallet rather than showing the connected wallets account?

And if I have two wallets, one with a position and one without. If I'm looking at my position -> switch to the wallet without an account we want to say on the position page rather than navigating to home?

I do like the idea of having all pages accessible, but I think many users are used to the app reflecting/ updating when switching the connected wallet

noahlitvin:
I was imagining the relevant functionality would just be disabled, you could navigate back to home from the top nav or 'account overview' link. I see where you're coming from though, maybe it's more natural to treat wallet change as like you've logged out and logged back in.

bachstatter:
> I was imagining the relevant functionality would just be disabled, you could navigate back to home from the top nav or 'account overview' link. I see where you're coming from though, maybe it's more natural to treat wallet change as like you've logged out and logged back in.

Yea I think changing a wallet if more like a logout login to me.. To get some real benefit of the not connected state I think we need to have a way to search/ add wallet or accounts too.

PR #520, opened 2022-11-23
gnosis and trust init

jmzwar:
Working for dapp Browser/injected with trust, and dapp browser for gnosis.


bachstatter (approved):
Looks good once the third party deps with `latest` has been changed to proper versions

PR #523, opened 2022-11-24
Burn max UI updates

bachstatter:
Screenshot 2022-11-24 at 3 45 08 pm


jmzwar (commented):
See comment on the function, I'd say it is much simpler to have badge variants as defined in the theme file. 

Also is the gray for the Debt Cleared badge background correct? Looks a tad light.

PR #525, opened 2022-11-25
Show value of dSNX

bachstatter:


v2/components/WalletBalances/WalletBalances.tsx
```diff
@@ -60,13 +60,13 @@ const WalletBalancesUi: React.FC<{
         />
         <StatBox
           label={t('staking-v2.wallet-balances.d-snx-value')}
-          amount={dSNXBalance === undefined ? undefined : formatNumberToUsd(dSNXBalance)}
+          amount={dSNXUSDBalance === undefined ? undefined : formatNumberToUsd(dSNXUSDBalance)}
```
jmzwar, line 63: think you could do && here

bachstatter: 0 wouldn't be formatted then

jmzwar: Ah bloody javascript

PR #538, opened 2022-11-30
V3 UI Build uplift

noisekit:
- [x] Include generated contracts in git
- [x] Align v3-contracts build with the rest of monorepo, build and package it, make ready for publish
- [x] Remove *duplicate* ts-deployments from v3/ui and use v3-contracts package instead
- [x] Uplift v3/ui build and update all the aliases to build from sources (same as v2/ui). Production-ready'ish
- [x] Ensure local web server works well with new build
- [x] Ensure versel app works
- [x] Update vercel config to skip full monorepo build (reduces deploy time to ~2min)

Local version seems to work same as previous one

image

Vercel deployment is OK too

image

One notable issue is this one, but it happens on master deployment too, so not a problem.
image

Rainbowkit CSS is back and wallet popups looking ok

image

image


v3/contracts/deployments/hardhat/Multicall.json
```diff
@@ -0,0 +1,424 @@
+{
```
bachstatter, line 1: Why do we have contracts for hardhard in this folder and in `src/hardhard`?

v3/ui/src/hooks/useCollateralTypes.ts
```diff
@@ -5,7 +5,7 @@ import { useRecoilState } from 'recoil';
 import { useContractReads, useProvider } from 'wagmi';
 import { chainIdState, collateralTypesState } from '../utils/state';
 import { localCollateralTypes } from '../utils/constants';
-import { abi as AggregatorABI } from '../../ts-deployments/aggregator';
+import { abi as AggregatorABI } from '@synthetixio/v3-contracts/build/hardhat/aggregator_snx.aggregator';
```
bachstatter, line 8: Should we be getting the ABI from hardhat? I would suggest mainnet, but since that doesn't exists maybe goerli makes the most sense?

bachstatter: Yea fair enought. I guess it should go into `v3/ui/src/utils/constants.ts` with a switch function or something

v3/contracts/deployments/hardhat/Multicall.json
bachstatter: ah make sense. Could maybe have a better name for the folder

v3/contracts/package.json
```diff
@@ -12,45 +21,17 @@
     "@typechain/ethers-v5": "^10.1.0",
     "@typechain/hardhat": "^6.1.3",
     "@types/node": "^18.8.5",
-    "cross-env": "^7.0.3",
```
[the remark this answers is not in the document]
bachstatter: I think `cross-env` make sure setting env variables work on windows machine. Dont think thats needed though. Looks like it wasn't use either way

0xjocke (approved):

PR #541, opened 2022-12-02
navbar

jmzwar:
Minor fixes


v2/components/Navigation/Navigation.tsx
```diff
@@ -211,6 +184,38 @@ export const NavigationUI = ({
             )}
           </Menu>
         </Center>
+        {}
```
bachstatter, line 187: Left a little extra `{}` here

jmzwar: doneskis

bachstatter (approved):

PR #542, opened 2022-12-02
Add test and fix claim button.

bachstatter:
- Should not be disabled when variant isn't success


jmzwar (approved):
u keep frontrunning my prs

bachstatter:
Pay more gas

bachstatter:
Ops there some progress bar things that is unrelated..

PR #543, opened 2022-12-02
Use latest foundry version

noisekit:


.circleci/config.yml
```diff
@@ -202,8 +202,7 @@ jobs:
           command: |-
             curl -L https://foundry.paradigm.xyz | bash
             source /root/.bashrc
-            # The latest version of foundry seems to be failing, locking it to the latest working one
-            foundryup -v nightly-15204ec245c11a521f5d48c8b79990cdc85f2198
+            foundryup
```
bachstatter, line 205: Normally I would prefer to lock to a version. But if they remove release versions I guess that's not a thing

bachstatter (approved):

PR #546, opened 2022-12-05
V3 type errors

bachstatter:
This PR fixes type errors:
Staking positions still loading
Screenshot 2022-12-05 at 2 23 08 pm
Screenshot 2022-12-05 at 2 22 59 pm

There's still some figuring out around how we interact with contracts. The generic `useContractRead` is not very type friendly. It is suppose to have typescript support if you use the latest wagmi version + `abi as const`. I did play around a little trying to make that work but wagmi told me our ABI was invalid. Tried with both json and "string" abi versions. 

It's also a bit sad that they way they structured their react query means that `select` cant be used.

Leaving it for now, but keen to hear opinions from v3 peeps that's been working with wagmi


bachstatter:
> Makes sense to me (I'm happy as long as it's green tbh)
Just need to figure out this line:
https://github.com/Synthetixio/js-monorepo/blob/v3-type-errors/v3/ui/src/hooks/useCollateralTypes.ts#L8
Wrote to you on discord

jmzwar (approved):

v3/ui/src/hooks/useNavigateWithChain.ts
```diff
@@ -16,7 +16,7 @@ export const useNavigateWithChain = () => {
     navigate(
       {
         ...to,
-        search: `?${createSearchParams({ chain: chainName })}`,
+        search: `?${createSearchParams({ chain: chainName || '' })}`,
```
sunnyvempati, line 19: prolly better to just remove the search param entirely if its empty string:
```
   ...{Boolean(chainName) ? { search: createSearchParams({ chain: chainName }) } : {}
```

sunnyvempati (commented):
i think the new wagmi versions have much better type support (i think you mentioned this); even if we can specify the type through a generic it would be better; not sure what the state of wagmi is but worth checking out and upgrading sooner rather than later

bachstatter:
> i think the new wagmi versions have much better type support (i think you mentioned this); even if we can specify the type through a generic it would be better; not sure what the state of wagmi is but worth checking out and upgrading sooner rather than later

Yea I did try the latest version but it thought consider our abi's to be invalid. In general, if we have typechain types and hook for getting `CoreProxy` we should have all methods typed. The methods can be called directly `CoreProxy.myMethod`, or `CoreProxy.populateTransaction.myMethod` for a multicall with type safety.

bachstatter:
Current CI error is happening on master too if you run `cannon-build` in v3/contracts. Pining in the tooling team channel. Noah said that Mati was aware of an issue.

bachstatter:
There's lots of work around here that will disappear once cannon have a release (hopefully tomorrow). I think should also work towards having a cannon file for downloading synthetix artefacts defined in synthetix-v3 instead. 

At least now we have the most update to date contracts. There is still an issue in that you need to run: 
```
npx cannon @ipfs:QmesjmjbXHj834SfavKuvWFFmepP3KdDev8vk9Unkz5i3f
npx cannon @ipfs:QmXdABDDVKr6Xx6SUUf3hPkv5VF4BzUHCk1FA8RF5KzxWf
npx cannon synthetix:3.0.0-alpha.4
```
This should hopefully disappear too. It feels a little wrong having these hacks in CI. So I'm just gonna leave typechecking off a few more days. At least now we have all the contract artefacts up to date

bachstatter:
Actually maybe we never need to have cannon-build in CI since we are committing the src's

PR #547, opened 2022-12-05
liquidation threshold of 0.01 to claim

jmzwar:
Prevent this scenario
Screen Shot 2022-12-05 at 8 32 10 pm


bachstatter (changes requested):
Update tests too

bachstatter (approved):
Hmm it's just the e2e acting out, will have a look tomorrow

PR #562, opened 2022-12-11
Subgraph market snapshot

bachstatter:
This PR:
- Extract out market event handlers to a seperate file, and separately tested 
- Add a new MarketSnapshot Entity
- Make sure that a MarketSnapshot entity is created everytime someone deposit/withdraws from a market


v3/subgraphs/goerli/src/market.ts
```diff
@@ -0,0 +1,96 @@
...
+
+function createMarketSnapshot(
+  marketId: string,
+  timestamp: BigInt,
+  usdDeposited: BigDecimal,
+  usdWithdrawn: BigDecimal,
+  netIssuance: BigDecimal,
+  reportedDebt: BigDecimal
+): void {
+  const newMarketSnapshot = new MarketSnapshot(marketId.concat('-').concat(timestamp.toString()));
```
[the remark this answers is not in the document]
bachstatter: This would happen everytime someone withdrawals or deposit for a market. The marketID + Timestamp is a true identifier for this entity in my mind. We could add a block number too but not sure it adds any value.
Or did you have anything else in mind

v3/subgraphs/goerli/tests/market.test.ts
```diff
@@ -0,0 +1,112 @@
...
+} from './event-factories';
+
+describe('Market tests', () => {
+  beforeEach(() => {
+    clearStore();
+  });
+
+  test('handleMarketCreated', () => {
+    // Needs to be here because of Closures
+    const now = new Date(1668448739566).getTime();
```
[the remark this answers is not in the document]
bachstatter: Yea agree I changed the other market test to have simpler timestamps. Will change this too

bachstatter:
We know have a few more entities:
`Market`: Stay the same, will contain the most recent values
`MarketSnapshotByBlock`: Stores every market update for each block, if there's two market events in the same block they would be aggregate
`MarketSnapshotByDay`: Stores every market update for each day, all updates during the same UTC day will be aggregated
`MarketSnapshotByWeek`: Same thing but for week.

We talked about doing it more event source:y and have the handlers totally isolated.
But since we all depend on the most recent Market values in some cases I opted to just pass the updated Market.

v3/subgraphs/goerli/schema.graphql
```diff
@@ -28,8 +31,48 @@ type Market @entity {
...
+  usd_withdrawn: BigDecimal!
+  "withdrawn - deposited"
+  net_issuance: BigDecimal!
+  reported_debt: BigDecimal!
+}
+type MarketSnapshotByWeek @entity {
+  "marketId-YYYY-WeekNumber"
+  id: ID!
+  timestamp: BigInt!
+  block_number: BigInt!
```
fritzschoff, line 64: out of consistency I would add created_at and created_at_block.
Same with udpated_at + block
I guess you did that with timestamp and block_number, so maybe just rename it!

v3/subgraphs/goerli/tests/marketSnapshotByBlock.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  createMarketUsdDepositedEvent,
+  createMarketUsdWithdrawnEvent,
+} from './event-factories';
+describe('MarketSnapshotByBlock', () => {
+  beforeEach(() => {
+    clearStore();
+  });
+  test('Handles deposit and withdrawals', () => {
+    const now = 1;
+    const newMarketRegisteredEvent = createMarketCreatedEvent(1, address, now, now - 1000);
```
fritzschoff, line 27: so you creating a market with a block number -999 ? I mean technically not wrong but weird test fixture.
Would argue to do it the way we had it.

v3/subgraphs/goerli/tests/marketSnapshotByBlock.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      Address.fromString(address2),
+      BigInt.fromU64(300),
+      now + 2000,
+      now + 1000
+    );
+    const newUsdWithdrawnEvent1 = createMarketUsdWithdrawnEvent(
+      1,
+      Address.fromString(address2),
+      BigInt.fromU64(100),
+      now + 2000,
```
fritzschoff, line 61: same time as event above. Maybe you want to change that

fritzschoff (changes requested):

v3/subgraphs/goerli/schema.graphql
bachstatter: Hmm yea, I added these before when I was doing when unique marketsnapshot for every update. But now when there's aggregation I guess it does make sense to follow the same structure 👍

v3/subgraphs/goerli/tests/marketSnapshotByBlock.test.ts
bachstatter: I do want that. The timestamp will be the same if two events happen in the same block. And when that's the case we should be updating the existing one rather than creating a new

v3/subgraphs/goerli/schema.graphql
bachstatter: Fixed

v3/subgraphs/goerli/tests/marketSnapshotByBlock.test.ts
bachstatter: No minus blocks now

v3/subgraphs/goerli/schema.graphql
```diff
@@ -28,8 +31,54 @@ type Market @entity {
   reported_debt: BigDecimal!
 }
 
+type MarketSnapshotByBlock @entity {
+  "marketId-blocknumber"
+  id: ID!
+  updated_at: BigInt!
+  updated_at_block: BigInt!
+  created_at: BigInt!
+  created_at_block: BigInt!
```
fritzschoff, line 40: thanks!

v3/subgraphs/goerli/src/market.ts
```diff
@@ -0,0 +1,83 @@
...
+export function handleMarketUsdDeposited(event: MarketUsdDeposited): void {
+  const marketId = event.params.marketId.toString();
+  const market = Market.load(marketId);
+  if (market == null) {
+    log.error(
+      'Something went wrong, got a MarketUsdDeposited event for a market that doesnt exists ' +
+        event.params.marketId.toString(),
+      []
+    );
+    return;
```
fritzschoff, line 36: weird that you don't get a build error since the function is `void`. Maybe wrap the reset into an else? But up to your gusto

v3/subgraphs/goerli/src/market.ts
```diff
@@ -0,0 +1,83 @@
...
+  const contract = MarketManagerModule.bind(event.address);
+  const reportedDebt = contract.getMarketReportedDebt(event.params.marketId).toBigDecimal();
+  const netIssuance = market.net_issuance.minus(event.params.amount.toBigDecimal());
+  market.reported_debt = reportedDebt;
+  market.updated_at = event.block.timestamp;
+  market.updated_at_block = event.block.number;
+  market.net_issuance = netIssuance;
+  market.usd_deposited = usdDeposited;
+  market.save();
+  createMarketSnapshotByBlock(market);
```
fritzschoff, line 49: So this came to my mind, quote from the subgraph docs:
`This more elaborate way of storing many-to-many relationships will result in less data stored for the subgraph, and therefore to a subgraph that is often dramatically faster to index and to query.`
What I mean is that, if we have a lot of entities in the store, this will actually slow down the subgraph.

fritzschoff (changes requested):

v3/subgraphs/goerli/src/market.ts
fritzschoff: https://thegraph.com/docs/en/developing/creating-a-subgraph/#example-2
This is about many to many relations but the idea is the same with the entities

v3/subgraphs/goerli/src/market.ts
bachstatter: `return` nothing is `void`

v3/subgraphs/goerli/src/market.ts
bachstatter: Hmm yea I think having an entity like that does make sense when there's a many to many relationship. It's like a classic many-to-many table. But in our case each marketSnapshot only belong to one market so not sure it will help in this case.

And yea with these entities indexing will be slower, but there's no extra contract calls which perviously has been what really slows down indexing . And fetching of data will be faster,  say we want history over a month or even a week, if we only had data for every block it could get quite big and require pagination + plus frontend aggregation.  With the  aggregation in the subgraph we can calculate the APR based on a markets yearly performance without any problems.  Or show yearly charts etc.

v3/subgraphs/goerli/src/market.ts
bachstatter: I prefer an early return for this edge case, we could even throw since this would be an unexpected error. But an error log is quite visible in the subgraphs ui

bachstatter:
@fritzschoff  I requested review, hoping to get this merged tomorrow

v3/subgraphs/goerli/src/market.ts
fritzschoff: I see the need, I get that. But the example just points out that if you have many entities indexing will be slower or as the docs say `dramatically`. It is not about many to many relationships, this is there just because this statement was written in relationships section of the docs.

v3/subgraphs/goerli/src/market.ts
fritzschoff: ah okay cool, nvm then

v3/subgraphs/goerli/src/market.ts
bachstatter: Yea of coruse it gets slower, it requires a save in a db for each one. 
But we are following their recommendation for one to many relationship:
> For one-to-many relationships, the relationship should always be stored on the 'one' side, and the 'many' side should always be derived. Storing the relationship this way, rather than storing an array of entities on the 'many' side, will result in dramatically better performance for both indexing and querying the subgraph. In general, storing arrays of entities should be avoided as much as is practical.

fritzschoff (approved):

v3/subgraphs/goerli/src/market.ts
fritzschoff: We had a chat and decided that we keep the createMarketByBlock function for now but with a comment that if the subgraph is slowing down, we gonna remove that func.

bachstatter:
After a call with max yesterday I decided to remove MarketSnapshotByBlock. Two reasons:
-  He had been working with some subgraphs being slow and flakey due to entities with lots of item.
-  At the moment we dont have any requirement to show block granularity, and even in the future when we build charts daily might be fine

v3/subgraphs/goerli/tests/marketSnapshotByWeek.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    // @ts-ignore
+    const weekNumber2 = getISOWeekNumber(Date.parse('2022-10-11T12:43:00.000Z').getTime());
+    assert.stringEquals(weekNumber2.toString(), '41');
+    // @ts-ignore
+    const weekNumber3 = getISOWeekNumber(Date.parse('2022-12-12T10:26:04.485Z').getTime());
+    assert.stringEquals(weekNumber3.toString(), '50');
+  });
+
+  test('Handles deposit and withdrawals', () => {
+    // @ts-ignore
```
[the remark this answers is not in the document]
bachstatter: hmm yea this is for vscode.. Is webstorm smart enough to figure out that this is assembly script?

v3/subgraphs/goerli/schema.graphql
```diff
@@ -19,8 +19,42 @@ type Market @entity {
...
+  updated_at: BigInt!
+  updated_at_block: BigInt!
+  usd_deposited: BigDecimal!
+  usd_withdrawn: BigDecimal!
+  "withdrawn - deposited"
+  net_issuance: BigDecimal!
+  reported_debt: BigDecimal!
+}
+
+type MarketSnapshotByDay @entity {
```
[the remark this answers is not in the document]
bachstatter: I guess we could if we add a `aggregation_period` field. The id structure would be different depending on that field too then.  Do you think that's better?

v3/subgraphs/goerli/tests/marketSnapshotByWeek.test.ts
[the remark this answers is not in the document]
bachstatter: Most of the types are correct, there are just in some cases where we use global stuff that exists in assembly script it gets confused

v3/subgraphs/goerli/schema.graphql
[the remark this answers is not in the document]
bachstatter: hmm say I only want weekly snapshots for the last 6months. I cant query that without getting all the daily snapshots too if they dont have a field that differentiate them. 

It would also obv work fine to have it like we have it now

PR #569, opened 2022-12-14
Switch Subgraph to use a single CoreProxy datasource

noisekit:
As we should not tap into internal module implementations and only rely on Proxy modules - there is no longer need to have multiple data sources. Everything can be taken from the same CoreProxy contract

We can still sepearate event handlers (see `market.ts`)


bachstatter (approved):
Nice make sense

PR #571, opened 2022-12-15
Fix display amount in unflag and self-liquidate

jmzwar:
Main thing is to use liquidation penalty data to render correct amount

Screen Shot 2022-12-15 at 12 37 26 pm


bachstatter (approved):
Thanks, sorry for forgetting to replace hardcoded stuff.
A `yarn && yarn deps:fix && yarn deps:mismatched --fix`  should fix ci

PR #572, opened 2022-12-15
Fix pool configuration

0xjocke:
I started digging into an sync error on the subgraph that surfaced now when we have some more data added to the contracts:
image

This turned out to require a few more changes than expected.
Some key learning:
1. We should never (or rather **CANT**) set derived fields, the compiler **wont** catch this but it fails when live
2. It's not possible to read derived fields from events handlers.

Give this sorta annoying findings this PR is suppose to fix the `PoolConfigurationSet` event by:
- Making sure we dont set configurations on the `Pool` or the `Market`
- Add a new field, `market_ids` on the `Pool`, this is needed so we know what `MarketConfiguration`s to remove when the `PoolConfigurationSet` is triggered
(This is the recommended architecture when you need to read derived fields: https://github.com/graphprotocol/graph-tooling/issues/1093)
I also made some fairly large refactor breaking things into functions and putting them in its own module. So now it's quite clear to see what mutative operations is done on a  `PoolConfigurationSet` event. That is:
1. `deleteRemovedMarketConfigurations`
2. `updateExistingMarketConfigurations`
3. `createNewMarketConfigurations`
4. `updatePoolFields`

### DIFF
--- a/v3/subgraphs/goerli/src/core.ts
+++ b/v3/subgraphs/goerli/src/core.ts
@@ -6,7 +6,6 @@ import {
   Liquidation,
   PermissionGranted,
   PermissionRevoked,
-  PoolConfigurationSet,
   PoolCreated,
   PoolNameUpdated,
   PoolNominationRenounced,
@@ -27,7 +26,6 @@ import {
   AccountRewardsDistributor,
   CollateralType,
   Liquidation as LiquidationEntity,
-  MarketConfiguration,
   Pool,
   Position,
   RewardsClaimed,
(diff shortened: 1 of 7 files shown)


v3/subgraphs/goerli/src/core.ts
```diff
@@ -37,7 +34,7 @@ import {
   Vault,
   VaultLiquidation as VaultLiquidationEntity,
 } from '../generated/schema';
-import { BigDecimal, BigInt, Bytes, store } from '@graphprotocol/graph-ts';
+import { BigDecimal, BigInt, Bytes, log, store } from '@graphprotocol/graph-ts';
```
fritzschoff, line 37: remove log please

fritzschoff (approved):

noisekit (approved):
Looking good to me!

PR #576, opened 2022-12-15
@usecannon/cli@1.1.22

noisekit:
- `@usecannon/cli@1.1.22`
- Add `fetch:<network>` and `start:<network>` scripts
- Lock synthetix v3 version
- Include fetch-deployments for all networks into build to have CI check for contract changes (we already have `git diff` after `yarn build`)


bachstatter (approved):
Looking good!

bachstatter:
Just needs a deps fix

PR #582, opened 2022-12-20
Wagmi upgrade

noisekit:
- account created OK

image

- Because the number was in `123e-11` format it crashed app on ethers.parse. Switched to wei instead (proper FIX: decimals for targetCRatio calculation as now issuanceRatio is 18 and not 6)

image

- Now works OK:
image

- typecheck is disabled as I don't want to waste time on this now since contracts interactions are being refactored (type errors come from contract interactions)
- Reuse same react-query client for wagmi
- Fixed mixed usage of `ethers.utils.parseUnits` and custom `parseUnits`. Stick to custom only (it is now based on wei and can deal with numbers like `123e-12`.
- walk over all `parseUnits`/`formatValue` and specify decimals (either dynamic from collateral or static 18 for known values like `liquidationRatioD18`)

Minting snxUSD went OK (tx https://goerli.etherscan.io/tx/0xd1cc62831da7e63e70a122b97c17d3acb330f8040b6a1a7558aafecf159a49e4)

Screenshot 2022-12-28 at 16 23 10
Screenshot 2022-12-28 at 16 23 23
Screenshot 2022-12-28 at 16 24 26

### DIFF
--- a/v3/ui/src/components/accounts/Position/useManagePosition.ts
+++ b/v3/ui/src/components/accounts/Position/useManagePosition.ts
@@ -60,7 +60,7 @@ export const useManagePosition = ({
             poolId,
             collateral.tokenAddress,
             currentAmount.add(collateralChangeBN),
-            parseUnits(1, 0),
+            parseUnits(1, 18),
           ],
         }
       );
@@ -95,7 +95,7 @@ export const useManagePosition = ({
             poolId,
             collateral.tokenAddress,
             currentAmount.sub(collateralChangeBN),
-            parseUnits(1, 0),
+            parseUnits(1, 18),
           ],
         },
         {
--- a/v3/ui/src/hooks/useDeposit.ts
+++ b/v3/ui/src/hooks/useDeposit.ts
@@ -85,7 +85,7 @@ export const useDeposit = ({
           parseInt(Boolean(accountId) ? selectedPoolId : poolId || '0'),
           selectedCollateralType.tokenAddress,
           amountToDelegate || 0,
-          parseUnits(1, 0),
+          parseUnits(1, 18),
         ],
       },
     ];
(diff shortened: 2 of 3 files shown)


v3/ui/src/hooks/useMulticall.ts
```diff
@@ -42,61 +42,32 @@ export const useMulticall = (
...
+  let callFunc: string | undefined;
+  let callArgs: any[] | undefined;
+  if (calls.length && snxProxy) {
     if (calls.length === 1) {
       // direct call
       callContract = calls[0].contract;
       callFunc = calls[0].functionName;
       callArgs = calls[0].callArgs;
     } else if (calls.length > 1) {
-      if (calls.find((c) => c.contract.address !== snxProxy?.address)) {
```
bachstatter, line 60: I would expect there to be a reason for why we use external multicall rather than the proxys multicall? Maybe the proxy's multicall doesn't work for external contracts? Did you try staking with WETH as collateral?
Worth asking @sunnyvempati  or @noahlitvin

bachstatter: Hmm yea looks like WETH hasn't been configured to be allowed as collateral. But I think the initial version will have that, will ping the v3 channel. Either way the frontend is written in a way to handle WETH and I think this change might break that

v3/ui/src/components/accounts/Position/useManagePosition.ts
```diff
@@ -44,65 +44,64 @@ export const useManagePosition = (
...
           functionName: 'delegateCollateral',
           callArgs: [
-            position.accountId,
-            position.poolId,
-            position.collateral.tokenAddress,
+            accountId,
+            poolId,
+            collateral.tokenAddress,
             currentAmount.add(collateralChangeBN),
-            utils.parseEther('1'),
```
bachstatter, line 63: parseEther has 18 deciamls. So everywhere parseEther was used, we should be doing: `parseUnits(1, 18)`
See example:
https://codesandbox.io/s/blissful-scooby-votem3?file=/src/App.tsx

bachstatter (approved):
Plan to sort out multicall when ETH is added as a valid collateral type

PR #586, opened 2022-12-29
Multi-network support for subgraphs

noisekit:
After a lot of back and forth, had to go back to graph's own multi-network deployment mechanism. The idea of extracting parts of subgraph into packages almost worked. But the main problem is - reusing `generated/....` functions and entities. They are per-network and (without significant dev inconveniences) we cannot inject those dependencies into library methods. Not saying it is impossible, but it becomes very sad experience.

Unfortunately it is only working if contracts (abis) are exactly same for all networks. Also graph cli re-compiles new subgraph.yaml file for each network, so it has to be git-ignored

When we start developing new versions on testnets that are different from mainnets we may just copy and create separate package like `snx-v3/subgraphs-canary`. But at this point copy-pasting and supporting multiple copies is not feasible.

Deployment and sync went ok for optimism-goerli
image


bachstatter (approved):

v3/subgraphs/subgraph.base.yaml
```diff
@@ -6,12 +6,9 @@ schema:
 dataSources:
   - kind: ethereum
     name: CoreProxy
-    network: goerli
 
     source:
-      address: '0x0E9628026e53f4c805073d85554A87dBd2011268'
       abi: CoreProxy
-      startBlock: 7833258
```
bachstatter, line 14: Did you think about how to handle this? syncing from optimism's start block might be slow?

PR #592, opened 2023-01-03
Disabled wagmi persistent storage

noisekit:
No more cache and json stringification of non-plain objects

image

### DIFF
--- a/v3/ui/src/index.tsx
+++ b/v3/ui/src/index.tsx
@@ -40,9 +40,15 @@ const { connectors } = getDefaultWallets({
   chains,
 });
 
+const noop = () => null;
 const wagmiClient = createClient({
   queryClient,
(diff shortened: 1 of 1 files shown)


v3/ui/src/index.tsx
```diff
@@ -43,6 +44,8 @@ const { connectors } = getDefaultWallets({
 const wagmiClient = createClient({
   queryClient,
   autoConnect: true,
+  // @ts-ignore
+  storage: noopStorage,
```
bachstatter, line 48: This would not give us client side cache either :(
image

bachstatter: Ah as long as react queries memory cache is still used this is perfect

bachstatter (approved):

PR #595, opened 2023-01-04
Deprecate poolsState, replace with usePools hook

noisekit:
Add useBlockchain for chain-related hooks to avoid relying on wagmi.
Currently wrapping wagmi but later will have different implementation.

Looks like weth price is borked
image


v3/lib/usePools/usePools.ts
```diff
@@ -0,0 +1,23 @@
...
+  return useQuery({
+    queryKey: [network.name, 'pools'],
+    queryFn: async () => {
+      if (!CoreProxyContract) throw new Error('CoreProxy');
+      const [preferredPool, approvedPools] = await Promise.all([
+        CoreProxyContract.getPreferredPool(),
+        CoreProxyContract.getApprovedPools(),
+      ]);
+      return [preferredPool]
+        .concat(approvedPools.filter((id) => !id.eq(preferredPool)))
```
bachstatter, line 17: This logic looks a bit weird? Looks like we can skip fetching `Approved Pools`?

v3/lib/useBlockchain/useBlockchain.ts
```diff
@@ -0,0 +1,24 @@
...
+  return signer;
+}
+
+export function useNetwork() {
+  const provider = wagmi.useProvider();
+  return {
+    id: provider.network.chainId,
+    name: provider.network.name,
+  };
+}
```
bachstatter, line 24: This is great

bachstatter (approved):

v3/lib/usePools/usePools.ts
bachstatter: hmm yea I figured you didn't change the logic..  ah this logic was written with the assumption that the preferred pool can be in the approved pool list too

PR #603, opened 2023-01-05
Account id param

noisekit:
Things still holding up well

image

### DIFF
--- a/v3/ui/src/hooks/useGetPoolData.ts
+++ b/v3/ui/src/hooks/useGetPoolData.ts
@@ -110,7 +110,7 @@ const getPoolData = async (chainName: string, id: string) => {
 export const useGetPoolData = (poolId?: string) => {
   const network = useNetwork();
 
-  return useQuery<Pool | undefined>({
+  return useQuery({
     queryKey: [network.name, 'pool', { poolId }],
     queryFn: async () => {
       if (!network.name || !poolId) throw Error('OMG!');
--- a/v3/lib/useAccounts/useAccounts.ts
+++ b/v3/lib/useAccounts/useAccounts.ts
@@ -5,7 +5,7 @@ import { useAccount } from '@snx-v3/useBlockchain';
 export function useAccounts() {
   const { address } = useAccount();
   const { data: AccountProxy } = useAccountProxy();
-  return useQuery<string[]>({
+  return useQuery({
     queryKey: [{ AccountProxy: AccountProxy?.address }, 'accounts'],
     queryFn: async function () {
       if (!AccountProxy) throw new Error('Should be disabled');
--- a/v3/ui/src/hooks/useRewards.ts
+++ b/v3/ui/src/hooks/useRewards.ts
@@ -3,11 +3,6 @@ import { CollateralType } from '@snx-v3/useCollateralTypes';
 import { useNetwork } from '@snx-v3/useBlockchain';
 import { useCoreProxy } from '@snx-v3/useCoreProxy';
 
-interface RewardDistribution {
-  value: string;
-  distributor: string;
-}
-
 export const useRewards = ({
   accountId,
   poolId,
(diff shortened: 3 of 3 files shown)


v3/ui/src/components/accounts/Deposit/index.tsx
```diff
@@ -240,14 +243,14 @@ export const DepositForm: FC<Props> = ({ accountId, liquidityPositions = {}, ref
...
-export const Deposit: FC<Props> = ({ accountId, liquidityPositions = {}, refetch }) => {
+export const Deposit: FC<Props> = ({ liquidityPositions = {}, refetch }) => {
   const { data: collateralTypes } = useCollateralTypes();
   const { chain: activeChain } = useNetwork();
   const hasWalletConnected = Boolean(activeChain);
   return (
     <>
       {collateralTypes && collateralTypes.length > 0 ? (
-        <DepositForm {...{ accountId, liquidityPositions, refetch }} />
+        <DepositForm {...{ liquidityPositions, refetch }} />
```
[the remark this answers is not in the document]
bachstatter: Yea I was gonna move `liquidityPositions` but then the skeletons are defined in the parent component. So left it for now.. 

Also there might be some merge conflicts here

v3/ui/src/hooks/useGetPoolData.ts
```diff
@@ -110,19 +110,21 @@ const getPoolData = async (chainName: string, id: string) => {
 export const useGetPoolData = (poolId?: string) => {
   const network = useNetwork();
 
-  return useQuery({
+  return useQuery<Pool | undefined>({
```
bachstatter, line 113: I think there's some problems with specifying types like this to react query. Can remember now though. But it should infer without this?

v3/ui/src/hooks/useGetPoolData.ts
```diff
@@ -110,19 +110,21 @@ const getPoolData = async (chainName: string, id: string) => {
...
     queryKey: [network.name, 'pool', { poolId }],
     queryFn: async () => {
       if (!network.name || !poolId) throw Error('OMG!');
       const poolData = await getPoolData(network.name, poolId);
+      if (!poolData.data.pool) {
+        return undefined;
```
bachstatter, line 119: I think `null` is slightly better? The data is `undefined` when query hasn't been fetched. `null` means there were no pool with this id?

bachstatter: > Null is problematic from types prospective tho, cannot use bla? anymore and need to do more explicit checks
I think that use to be the case but not true anymore
image

bachstatter: >  essentially, downstream we would need to have separate checks for undefined, null and non-empty. We don't do it - but ideally we should.

I dont think we should change it to undefined, it just make it slightly worse. But yea it would be great to show a pool does not exist message on the pool page

v3/ui/src/hooks/useGetPoolData.ts
bachstatter: I'm happy to help figuring out if the infer types behaves weirdly, it really shouldn't. The problem with specifying types is that `select` cant be used. If we want to explicitly type queries we should give the generic 4 types
https://tkdodo.eu/blog/react-query-and-type-script#the-four-generics

But the recommended way is to not specify types at all:
https://tkdodo.eu/blog/react-query-and-type-script#type-inference

bachstatter: If it's working now just remove the explicit types then ready to merge

bachstatter (approved):

v3/ui/src/hooks/useDeposit.ts
```diff
@@ -52,7 +52,7 @@ export const useDeposit = ({
 
   const { wrap, balance: wrapEthBalance, isLoading: isWrapping } = useWrapEth();
 
-  const newAccountId = useMemo(() => Math.floor(Math.random() * 10000000000), []);
+  const newAccountId = useMemo(() => `${Math.floor(Math.random() * 10000000000)}`, []);
```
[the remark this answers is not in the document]
bachstatter: Feel free to change it. It's still fairly low change to get a duplicate and if that happens the worst thing that happens is the transaction fails.

PR #610, opened 2023-01-10
Refactor Teleporter to have a list of multiple transactions without global state

noisekit:
- Copied components into a package, no changes to unaffected useDeposit/useManagePosition for now for backward compatibility
- Fixed jest config

Visually nothing changed
image


bachstatter (approved):
Looks very clean. We do have a `txnReducer` which have the states slightly different. Would be good to align those somehow?  You can see how it's used here: https://github.com/Synthetixio/js-monorepo/blob/master/v3/lib/useApprove/useApprove.ts#L60

PR #612, opened 2023-01-10
Refactor and split Deposit, remove react-hook-form

noisekit:
Refactor `Deposit` to be a simple form with onChange/onSubmit and standard HTML Form Validity checks (with a couple of custom messages)

Extract a few components and libs along the way + refactor out wagmi and avoid unnecessary props drilling:
- `@snx-v3/Amount`
- `@snx-v3/Balance`
- `@snx-v3/CollateralTypeSelector`
- `@snx-v3/PoolSelector`
- `currency` moved to `@snx-v3/format` (we should reuse our default formatter from v2 and deprecate this one)

_ 2023-01-10 at 12 23 39
_ 2023-01-10 at 12 24 08
_ 2023-01-10 at 12 24 18

Pool selector:

image


v3/components/PoolSelector/PoolSelector.tsx
```diff
@@ -0,0 +1,33 @@
...
+  setPoolId,
+}: {
+  poolId: string;
+  setPoolId: (poolId: string) => void;
+}) {
+  const { data: pools = [] } = usePools();
+  const [value, setValue] = useState(poolId);
+  useEffect(() => {
+    setValue(poolId);
+  }, [poolId]);
```
bachstatter, line 17: Could you just rely on `poolId` and `setPoolId`?

bachstatter: Hmm yea I guess we can start making these decisions when we implement the new designs

bachstatter (approved):

PR #613, opened 2023-01-10
Fix types for Amount

noisekit:
WeiSource is better

### DIFF
--- a/v3/components/Amount/Amount.tsx
+++ b/v3/components/Amount/Amount.tsx
@@ -1,7 +1,7 @@
 import { Tooltip } from '@chakra-ui/react';
-import { BigNumber } from 'ethers';
 import { useMemo } from 'react';
-import { currency, parseUnits, formatValue } from '@snx-v3/format';
+import { currency, formatValue, parseUnits } from '@snx-v3/format';
+import { WeiSource } from '@synthetixio/wei';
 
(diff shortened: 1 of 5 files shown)


v3/components/Amount/Amount.tsx
```diff
@@ -1,15 +1,15 @@
 import { Tooltip } from '@chakra-ui/react';
-import { BigNumber } from 'ethers';
 import { useMemo } from 'react';
-import { currency, parseUnits, formatValue } from '@snx-v3/format';
+import { currency, formatValue, parseUnits } from '@snx-v3/format';
```
bachstatter, line 3: `formatValue` expects bn and not wei source?

bachstatter: What happens if I pass `Wei` to as value
`const parsedValue = parseUnits(value)`

bachstatter: The patterns we used in v2, is that more pure uI components always gets passed number. Then format as needed. This worked better with storybook too which struggled with BigNumber/ Wei

bachstatter: Hmm yea, if `parseUnits` can handle `Wei` I think we can merge this for now

bachstatter (approved):

PR #618, opened 2023-01-12
Fix account crash

bachstatter:
- Remove Account page. Replaced by home screen
- Update Depositing Nav and AccountsSelector to new route strucutre


bachstatter:
@noisekit  this might be out of date with your changes to routing. But removal of that account component still valid

PR #623, opened 2023-01-13
Deposit modal

noisekit:
- Wrapping removed and re-added but smarter (it is no longer optional if weth is required, but we wrap only what is necessary and no more)!
- Many small changes around collateral/deposit form/etc

image

image

image

Teleporter still works (sort of)
image

### DIFF
--- a/v3/ui/src/components/accounts/Deposit/DepositForm.tsx
+++ b/v3/ui/src/components/accounts/Deposit/DepositForm.tsx
@@ -22,19 +22,25 @@ export function DepositForm() {
 
   const inputRef = useRef<HTMLInputElement | null>(null);
   const navigate = useNavigate();
-  const [amount, setAmount] = useState('');
+  const [inputAmount, setInputAmount] = useState('');
+  const [amount, setAmount] = useState(wei(0));
   const [activeBadge, setActiveBadge] = useState(0);
 
   const balanceData = useTokenBalance(collateralType?.tokenAddress);
 
   const [isOpenDeposit, setIsOpenDeposit] = useState(false);
-  const onSubmit = useCallback(async (e: FormEvent) => {
-    e.preventDefault();
-    const form = e.target as HTMLFormElement;
-    if (form.reportValidity()) {
+  const onSubmit = useCallback(
+    async (e: FormEvent) => {
+      e.preventDefault();
+      const form = e.target as HTMLFormElement;
+      if (!form.reportValidity()) {
+        return;
+      }
+      setAmount(wei(inputAmount));
       setIsOpenDeposit(true);
-    }
-  }, []);
+    },
+    [inputAmount]
+  );
 
   const onChangeCollateral = useCallback(
     (collateralSymbol: string) => {
@@ -45,7 +51,8 @@ export function DepositForm() {
         return;
       }
       setActiveBadge(0);
-      setAmount('');
+      setAmount(wei(0));
+      setInputAmount('');
       inputRef.current?.focus();
       navigate({
         pathname: generatePath('/deposit/:collateralSymbol/:poolId', {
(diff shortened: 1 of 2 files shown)


v3/components/TransactionReview/TransactionStatus.ts
```diff
@@ -1 +1,6 @@
-export type TransactionStatus = 'idle' | 'processing' | 'failed' | 'completed';
+export type TransactionStatus = {
+  failed?: boolean;
+  disabled?: boolean;
+  loading?: boolean;
+  success?: boolean;
+};
```
bachstatter, line 6: I would slightly prefer the old structure of this. Then you have type safety for accidentally setting two of the flags to true

v3/ui/src/components/accounts/Deposit/DepositModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    setFailed(false);
+    setProcessing(true);
+
+    setStep(1);
+    if (requireApproval) {
+      try {
+        await approve(Boolean(infiniteApproval));
+        await refetchAllowance();
```
bachstatter, line 216: We dont really have to await this. The next transaction will go through since the allowance have been set. Maybe move it to when all transactions are finished? Or just not await

v3/components/TransactionReview/TransactionStatus.ts
jmzwar: Agree with Joey on this one soz

v3/components/TransactionReview/statusColor.ts
```diff
@@ -1,14 +1,17 @@
 import { TransactionStatus } from './TransactionStatus';
 
 export function statusColor(status: TransactionStatus): string {
-  switch (status) {
-    case 'idle':
+  switch (true) {
```
jmzwar, line 4: I believe we can create variants for this? I.e error variant, disabled variant etc

v3/components/TransactionReview/TransactionStatus.ts
jmzwar: I might be missing some context here but I don't understand how a transaction can be in success or disabled state at the same time.

v3/ui/src/components/accounts/Deposit/DepositModal.tsx
```diff
@@ -0,0 +1,313 @@
...
+  accountId: string;
+  amount: string;
+  poolId: string;
+  collateralType: CollateralType;
+  isOpen: boolean;
+  setIsOpen: (isOpen: boolean) => void;
+}) {
+  const navigate = useNavigate();
+
+  const toast = useToast({
```
jmzwar, line 43: Nice - might make sense in the future to extract these toast settings out to reuse.

v3/components/TransactionReview/TransactionStatus.ts
jmzwar: Can you link me to the context in figma?

jmzwar: > > I might be missing some context here but I don't understand how a transaction can be in success or disabled state at the same time.
> 
> It is not a transaction. It is an operation within a list of multiple operations. It usually includes a transaction but also has way more to it.

Hmm ok so it has depth to it, in that case would it be worth renaming to something more verbose?

v3/ui/src/components/accounts/Deposit/DepositModal.tsx
```diff
@@ -0,0 +1,313 @@
...
+  return (
+    <Modal
+      size="lg"
+      isOpen={isOpen}
+      onClose={() => {
+        setStep(0);
+        setCompleted(false);
+        setFailed(false);
+        setProcessing(false);
+        setIsOpen(false);
```
jmzwar, line 257: Oof

jmzwar: Here i'd reach for JZ's grandpa approach of using a reducer

v3/components/TransactionReview/TransactionStatus.ts
jmzwar: I was thinking Multistep would be good to delineate it from a Single transaction status

v3/ui/src/components/accounts/Deposit/DepositModal.tsx
```diff
@@ -0,0 +1,359 @@
...
+            {(() => {
+              switch (true) {
+                case failed:
+                  return 'Retry';
+                case processing:
+                  return 'Processing...';
+                case completed:
+                  return 'Done';
+                default:
+                  return 'Start';
```
jmzwar, line 351: Inline jsx switch is also an option here

```
{
        {
          'failed': 'Retry',
          'processing': 'Processing...',
          'completed': 'Done',
        } [param] || 'Start'
}

File: v3/ui/src/components/accounts/Deposit/DepositModal.tsx (hunk 2 above)
Comment (bachstatter) [reply v3/ui/src/components/accounts/Deposit/DepositModal.tsx:246]: Maybe send it of at the same time as the final in a Promise.all then? Unnecessary to wait for an async action before starting another async action?

File: v3/ui/src/components/accounts/Deposit/DepositForm.tsx [hunk 7]
```diff
@@ -94,17 +101,13 @@ export function DepositForm({
...
                 _focus={{ boxShadow: 'none !important' }}
                 _placeholder={{ color: 'whiteAlpha.700' }}
-                min="0"
                 required
                 value={numberWithCommas(amount)}
                 onChange={(e) => {
                   const value = e.target.value.replaceAll(',', '');
-                  if (!/^[0-9]*(\.[0-9]{0,2})?$/.test(value)) {
-                    return;
-                  }
```
bachstatter, line 104: With this removed we crash if user types a non number character

v3/components/TransactionReview/TransactionStatus.ts
bachstatter: This is not a biggie for me. But it feels like if it can be two things at the same time they could be separated. Def not a blocker though

v3/ui/src/components/accounts/Deposit/DepositModal.tsx
bachstatter: If `await approve(Boolean(infiniteApproval));`  doesn't throw we will have approval updated on the contract. The nicest pattern for the user would probably be to invalidate/ update the allowance cache when approval succeeds.
Happy to leave this for now, but it is already a fairly long process to deposit so would be nice to make it as quick as possible

v3/ui/src/components/accounts/Deposit/DepositForm.tsx
bachstatter: It still crashes for me, I think it make sense not to update the state if user types a non number. Since there are many places where we assume `amount: string` can be converted to a number

bachstatter: I prefer not having any characters in our number fields either way. This is how the number field in manage position will work, since we will update calculation as user types. It also how it works for on 1inch, UNI and AAVE so users should be use to it

bachstatter (approved):

PR #624, opened 2023-01-13
Whitelist oracle manager proxy

bachstatter:


bachstatter:
Hmm yea.. And according to DB an upgrade coming. Maybe that resolves it

PR #627, opened 2023-01-16
ETH <-> WETH separation

noisekit:
Few extra fixes:
- Account can be created dynamically so don't require it for DepositModal
- Ensure accountId param on Deposit page when we know it
- wei -> bn args for contracts on DepositModal
- refetch ETH balance after deposit
- Make sure we have accountId in params when being on Deposit page

_ 2023-01-16 at 11 08 05
_ 2023-01-16 at 11 51 50
_ 2023-01-16 at 11 49 28
_ 2023-01-16 at 11 49 36

### DIFF
--- a/v3/ui/package.json
+++ b/v3/ui/package.json
@@ -32,6 +32,7 @@
     "@snx-v3/useBlockchain": "workspace:*",
     "@snx-v3/useCollateralTypes": "workspace:*",
     "@snx-v3/useCoreProxy": "workspace:*",
+    "@snx-v3/useEthBalance": "workspace:*",
     "@snx-v3/useExternalMulticall": "workspace:*",
     "@snx-v3/useLiquidityPosition": "workspace:*",
     "@snx-v3/useLiquidityPositions": "workspace:*",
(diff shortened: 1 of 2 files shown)


v3/ui/src/components/accounts/LiquidityPositions/LiquidityPosition.tsx
```diff
@@ -19,7 +19,7 @@ export const LiquidityPosition: FC<Props> = ({ position }) => {
           <Text fontSize="xs" opacity="0.66" mt="1">
             <Amount
               value={position.collateralAmount}
-              suffix={`${position.collateralType.symbol.toUpperCase()} `}
+              suffix={`${position.collateralType.symbol} `}
```
bachstatter, line 22: Here we might wont to display ETH rather than WETH

bachstatter: Yea I now it is WETH, but we do want to display it as ETH everywhere (except on the deposit actions). The oracle price will be from and ETH oracle too.. Doesn't have to happen in this PR, but should at some point

bachstatter (approved):

PR #630, opened 2023-01-17
Simplify use liquidity position

noisekit:
Part of Manage Position refactoring, just splitting into a few very small PRs with limited scope for quick and easy review.


v3/lib/useLiquidityPosition/useLiquidityPosition.ts
```diff
@@ -62,11 +61,11 @@ export const loadPosition = async ({
 };
 
 export const useLiquidityPosition = ({
-  collateral,
+  tokenAddress,
```
bachstatter, line 64: collateralAddress slightly better?

bachstatter (approved):
Comment no biggie

PR #637, opened 2023-01-18
V2 burn fixes

bachstatter:
- Dont round numbers for preset.
- Send whole usd balance when `debtBalance` preset is active (see comment)
- Allow 18 dp in input field
image


jmzwar:
> How dis it happen I fixed exactly same thing today but for v3 😆

Are we sharing any logic across mint/burn in v2/v3?

Might make sense to extract out

bachstatter:
> > How dis it happen I fixed exactly same thing today but for v3 😆
> 
> Are we sharing any logic across mint/burn in v2/v3?
> 
> Might make sense to extract out

I copied and pasted a fair amount when I did the deposit page

PR #638, opened 2023-01-18
Validate position and number input

noisekit:
- Rename `useValidatePosition` to `validatePosition`
- Make `@snx-v3/NumberInput` to be somewhat more usable and avoid altering input too much
- Convert `NumberInput` to wei and update all related code to wei (wei contagion)
- Reduce precision-based errors by removing parseUnits/format from all the places, except the very last. Format only when doing UI rendering, avoid conversion of data mid-way, pass wei throughout without any conversions

Few screenshots (no more funkiness with numbers in the UI)

image

image

image

This one is weird. It expects debt repayment at the same time, so `isValid` is false. I did not change the behaviour here an we will have it different anyway.

image

### DIFF
--- a/v3/ui/src/pages/AccountPositionPage/Manage/Custom.tsx
+++ b/v3/ui/src/pages/AccountPositionPage/Manage/Custom.tsx
@@ -16,6 +16,7 @@ import { CollateralType } from '@snx-v3/useCollateralTypes';
 import { Balance } from '@snx-v3/Balance';
 import { NumberInput } from '@snx-v3/NumberInput';
 import { Wei } from '@synthetixio/wei';
+import { formatPercent } from '@snx-v2/formatters';
 
 export function Custom({
   collateral,
@@ -86,7 +87,11 @@ export function Custom({
               <Box>
                 <Text fontSize="xs">
                   Max Mint: ${currency(maxDebt)}
-                  <Tooltip label="You can't mint snxUSD that takes your C-Ratio below the target c-ratio of 300%.">
+                  <Tooltip
+                    label={`You can't mint snxUSD that takes your C-Ratio below the issuance ratio of ${formatPercent(
+                      collateral.issuanceRatioD18.toNumber()
+                    )}.`}
+                  >
                     <QuestionOutlineIcon transform="translateY(-1.5px)" ml="1" />
                   </Tooltip>
                   {maxDebt.eq(0) ? null : (


v3/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+  useEffect(() => {
+    if (!onChange) {
+      return;
+    }
+    const t = setTimeout(() => {
+      const nextValue = wei(inputValue.replaceAll(',', '') || 0);
+      if (value.eq(nextValue)) {
+        return;
+      }
+      if (max === undefined) {
```
jmzwar, line 31: This is a tad confusing - might make sense to add some comments

bachstatter: hmm yea. If I'm reading this right:
If a user adds a number bigger than max we will reset it to max 300ms later?
I dont think we want this, since we want to be able to show different values affect the c-ratio even when the user doesn't have enough balance.

v3/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+      if (value.eq(0)) {
+        return setInputValue('');
+      }
+      // Cleanup trailing precision zeroes
+      const float = parseFloat(value.toString());
+      if (float === value.toNumber()) {
+        return setInputValue(`${float}`);
+      }
+      return setInputValue(value.toString());
+    }, 100);
```
bachstatter, line 53: Should this be done in an event handler instead?

v3/components/NumberInput/NumberInput.tsx
jmzwar: The logic within the useEffect I mean

v3/ui/src/pages/AccountPositionPage/Manage/Mint.tsx
```diff
@@ -25,15 +29,17 @@ export function Mint({
...
         <Flex alignItems="center">
           <Box>
             <Text fontSize="xs">
-              Max Mint: ${currency(max.toNumber())}
+              Max Mint: ${currency(max)}
               <Tooltip
                 color="white"
-                label="You can't mint snxUSD that takes your C-Ratio below the target c-ratio of 300%."
+                label={`You can't mint snxUSD that takes your C-Ratio below the target c-ratio of ${formatPercent(
+                  collateral.liquidationRatioD18.toNumber()
```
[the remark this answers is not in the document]
bachstatter: This should say your C-Ratio below the **issuance** c-ratio of {collateral.issuanceRatio}

v3/components/NumberInput/NumberInput.tsx
bachstatter: > That was initial implementation I haven't changed. It was picking `Math.max(balance, input)` Same for all other inputs.
> 
> Do you want me to remove it? I'm all for it. Really don't like it myself.

Lets remove it. This will probs change a bit soon either way

bachstatter (approved):

PR #644, opened 2023-01-20
V3 contracts update

noisekit:
- cannon `2.0.0-alpha.7`
- contracts `3.0.0-alpha.7`
- refactor contracts `codegen` to work out breaking changes
- update app code to renamed contracts


bachstatter (approved):
Looks good.  Can you confirm that alpha.8 isn't out? We could also updgrade the CoreProxy address for the subgraphs, I think that's still manual?

PR #645, opened 2023-01-22
Feature/v3 header

jmzwar:
Screen Shot 2023-01-22 at 8 26 52 pm
Screen Shot 2023-01-22 at 8 26 58 pm


v3/components/icons/ChevronDown/ChevronDown.tsx
```diff
@@ -0,0 +1,12 @@
+import { Icon, IconProps } from '@chakra-ui/react';
```
bachstatter, line 1: I think chakra icons might have the chevrons

v3/ui/src/components/NetworkController.tsx
```diff
@@ -1,13 +1,40 @@
...
+  EthereumIcon,
+  OptimismIcon,
+  FailedIcon,
+  ChevronUp,
+  ChevronDown,
+  WalletIcon,
+} from '@snx-v3/icons';
 import { useAccount } from '@snx-v3/useBlockchain';
 
+import { switchNetwork } from '@wagmi/core';
```
bachstatter, line 13: https://wagmi.sh/react/hooks/useSwitchNetwork I think we should use this instead, then we dont have to add wagmi/core

v3/ui/package.json
```diff
@@ -54,6 +54,7 @@
     "@synthetixio/wei": "workspace:*",
     "@tanstack/react-query-devtools": "^4.3.5",
     "@wagmi/chains": "^0.1.5",
+    "@wagmi/core": "latest",
```
bachstatter, line 57: This should not be latest. But also I dont think we need it. See comment about swtich netowrk

bachstatter:
It should also display the account when you have a position, so make sure you do some testing with an account too

bachstatter (approved):

v3/ui/package.json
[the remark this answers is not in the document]
jmzwar: removed

v3/ui/src/components/NetworkController.tsx
jmzwar: Ta - done

PR #649, opened 2023-01-23
Manage page

bachstatter:
- Create a new temp manage page, until mange flows been build.  Same URL as current but starting with `/new/`
- Add some boilerplate so it's easier to split up things up between us
image


jmzwar (approved):
Border box can probably be some custom style defined in the theme but otherwise good

bachstatter:
Hmm yea I guess it could. It's used on many pages, but can tackle it in another PR

PR #654, opened 2023-01-24
Deposit form wei

bachstatter:
Styles NumberInput and uses it in deposit form.
Also use container pattern for Deposit Form

### DIFF
--- a/v3/ui/src/components/accounts/Deposit/DepositForm.tsx
+++ b/v3/ui/src/components/accounts/Deposit/DepositForm.tsx
@@ -99,10 +99,19 @@ export function DepositForm(props: { staticCollateral?: boolean }) {
         <Box borderWidth="1px" borderColor="gray.900" borderRadius="base" p={2}>
           <Flex justifyContent="space-between">
             <Flex alignItems="center">
-              <CollateralTypeSelector
-                collateralSymbol={params.collateralSymbol}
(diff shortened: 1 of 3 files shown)


v3/ui/src/components/accounts/Deposit/DepositForm.tsx
```diff
@@ -1,44 +1,55 @@
...
-      return tokenBalance.data;
+      return tokenBalance;
     }
-    if (!tokenBalance.data || !ethBalance.data) {
+    if (!tokenBalance || !ethBalance) {
       return undefined;
     }
-    return tokenBalance.data.add(ethBalance.data);
-  }, [collateralType?.symbol, tokenBalance.data, ethBalance.data]);
+    return tokenBalance.add(ethBalance);
```
jmzwar, line 51: Hmm perhaps this is where max is causing an issue? If we wrap as part of the multicall..

jmzwar (approved):

v3/ui/src/components/accounts/Deposit/DepositForm.tsx
bachstatter: Hmm max for ETH is annoying... cause if you do max on the first deposit call, you wont have any ETH left to delegate to the pool.
Need to discuss how we want that to work then fix

PR #668, opened 2023-01-30
Use deposit

bachstatter:
Added a `useDeposit` hook that use `CoreProxy.multicall` without wagmi

### DIFF
--- a/v3/ui/src/components/accounts/Deposit/DepositModal.tsx
+++ b/v3/ui/src/components/accounts/Deposit/DepositModal.tsx
@@ -124,8 +124,7 @@ export const DepositModal: DepositModalProps = ({
           title: Boolean(accountId)
             ? 'Depositing your collateral'
             : 'Creating your account and depositing collateral',
-          description:
-            "You'll be redirected to the manage position page once the transaction is confirmed.",
+          description: '',
           status: 'info',
           isClosable: true,
           duration: 9000,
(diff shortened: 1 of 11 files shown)


v3/ui/src/components/accounts/Deposit/DepositModal.tsx
```diff
@@ -255,7 +221,10 @@ export const DepositModal: DepositModalProps = ({
 
     setStep('deposit');
     try {
-      await multiTxn.exec();
+      // Once the previous `refetchAllowance` call finish, execDeposit gets called before useDeposit hook is enabled.
+      // If we enable the call before we have enough allowance, the estimate gas call will fail :(
+      await new Promise((res) => setTimeout(res, 1000));
```
bachstatter, line 226: This is obv not nice, but I couldn't figure out a nice way to fix it.. Happy to talk through this

jmzwar: Hmm - why is it necessary exactly?

bachstatter: Hmm the comment wasn't good enough I guess...
The `useDeposit` will try to estimate gas directly by calling populating the transaction, but it cant do that until we have approval since the tx will fail.
At this point in the code approval has just been updated, but the event loop will call `exec` before the `useDeposit` rerenders with the new approval.  And when no gas has been estimated and exec is called it will exit early..

With the `wait` the hook will have time to rerender

bachstatter: Hmm yea the only solution I can think of would require one extra state variable + useEffect.

bachstatter: hm actually, I can rely on the step state, so could solve it with just a useEffect

jmzwar: Can you abstract it away so it isn't in the component?

v3/ui/src/components/accounts/Deposit/DepositModal.tsx
```diff
@@ -276,9 +223,30 @@ export const DepositModal: DepositModalProps = ({
     approve,
     infiniteApproval,
     refetchAllowance,
-    multiTxn,
   ]);
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
bachstatter: useEffect removed now

jmzwar (approved):

PR #672, opened 2023-02-01
Update contracts and cannon

noisekit:
- contracts `3.0.0-alpha.8`
- cannon ~`2.0.2`~ `2.0.4`
- No more difference in subgraph config (remove copying of `subgraph.yaml`)
- Make sure we build subgraph in CI and check diffs
- Updated subgraph tests for new contract


bachstatter (approved):
Nice, things getting cleaner

PR #682, opened 2023-02-03
Generate random wallets for each e2e test

noisekit:
- fixed `useNetwork` and `useProvider` to work around wagmi inconsistencies
- updated few places in the code to use correct network (typechecked)

## custom command to generate wallet

it returns `ethers.Wallet` instance if we want to run some tasks for that address/pk
but also stores wallet instance in custom namespace (defaults to `@wallet` but can be redefined)

image

## signes txns with impersonation

image

## for UI it mimics normal wallet

image


bachstatter (approved):
Nice!!

PR #693, opened 2023-02-07
E2E creates new account with first deposit of SNX

noisekit:
- also skip WETH deposit test (until collateral is enabled again)

_ 2023-02-07 at 11 41 16

### DIFF
--- a/v3/cypress/cypress/e2e/New Account - Deposit SNX.e2e.js
+++ b/v3/cypress/cypress/e2e/New Account - Deposit SNX.e2e.js
@@ -1,0 +1,51 @@
+import { generatePath } from 'react-router-dom';
+
+it('creates new account with first deposit of SNX', () => {
+  cy.connectWallet().then(({ address }) => {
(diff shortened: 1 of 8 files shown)


v3/cypress/cypress/e2e/New Account - Deposit SNX.e2e.js
```diff
@@ -0,0 +1,51 @@
...
+    .should('include.text', 'Done')
+    .and('be.enabled')
+    .click();
+
+  cy.get('[data-testid="deposit modal"]').should('not.exist');
+
+  cy.location('pathname').should('include', 'accounts').should('include', 'positions');
+
+  cy.get('[data-testid="current account id"]').then((element) => {
+    const accountId = element.attr('data-accountId');
```
bachstatter, line 42: I know this isn't really relavant to this change, but react is complaining about `data-accountId` it doesn't like case-sensitive data attributes

bachstatter (approved):

PR #694, opened 2023-02-07
E2E Manage position: borrow

noisekit:
image

### DIFF
--- a/v3/cypress/cypress/tasks/delegateCollateral.js
+++ b/v3/cypress/cypress/tasks/delegateCollateral.js
@@ -1,0 +1,23 @@
+import { ethers } from 'ethers';
+import * as CoreProxy from '@synthetixio/v3-contracts/src/goerli/CoreProxy';
+import { getCollateralConfig } from './getCollateralConfig';
+
+export async function delegateCollateral({ privateKey, accountId, symbol, amount, poolId }) {
+  const config = await getCollateralConfig(symbol);
+  const provider = new ethers.providers.JsonRpcProvider('http://127.0.0.1:8545');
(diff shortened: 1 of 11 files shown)


v3/cypress/cypress/tasks/delegateCollateral.js
```diff
@@ -0,0 +1,23 @@
...
+  const wallet = new ethers.Wallet(privateKey, provider);
+  console.log('delegateCollateral', { address: wallet.address, accountId, symbol, amount, poolId });
+
+  const coreProxy = new ethers.Contract(CoreProxy.address, CoreProxy.abi, wallet);
+
+  const tx = await coreProxy.delegateCollateral(
+    ethers.BigNumber.from(accountId),
+    ethers.BigNumber.from(poolId),
+    config.tokenAddress,
+    ethers.utils.parseEther(`${amount}`),
```
bachstatter, line 17: I dont think we need any change since we are using fresh wallets + account. But worth mentioning that if you have any delegatedCollateral already this will add to that position rather than replacing

v3/ui/src/pages/Manage/ManageStats.tsx
```diff
@@ -80,7 +84,7 @@ export const ManageStatsUi: FC<{
         </Flex>
       </BorderBox>
       <BorderBox px={4} py={2} display="flex" gap={1} flexDirection="column">
-        <Text color="gray.500" fontSize="sm">
+        <Text color="gray.500" fontSize="sm" data-testid="manage stats c-ratio">
```
bachstatter, line 87: I think this data-attr is unused

bachstatter (approved):

jmzwar (approved):

PR #698, opened 2023-02-09
E2E Manage position: repay

noisekit:
- Updated `NumberInput` to set and reset validity independently of input (`max` value can come delayed, without actual input updates)
- Disabled submit button until we get `max`, `balance` and `debt` populated

image

image

### DIFF
--- a/v3/components/NumberInput/NumberInput.tsx
+++ b/v3/components/NumberInput/NumberInput.tsx
@@ -7,6 +7,8 @@ export interface NumberInputProps extends InputProps {
   'data-max'?: string;
 }
 
+export const NUMBER_REGEX = /^([0-9]*[.])?[0-9]{0,18}$/;
+
 export function NumberInput({
   value,
   onChange,
@@ -23,9 +25,11 @@ export function NumberInput({
   const onInputChange = useCallback(
     (e: ChangeEvent<HTMLInputElement>) => {
       setInputValue(e.target.value);
-      e.target.setCustomValidity('');
       if (!onChange) {
         // Could be a read-only input
+        return;
+      }
+      if (!NUMBER_REGEX.test(`${e.target.value}`)) {
         return;
       }
       let nextValue = value;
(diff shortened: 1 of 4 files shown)


v3/components/NumberInput/NumberInput.tsx
```diff
@@ -34,18 +34,25 @@ export function NumberInput({
...
 
+  const ref = useRef<HTMLInputElement>(null);
+  useEffect(() => {
+    if (!ref.current) {
+      return;
+    }
+    if (max && max.gte(0) && value && value.gt(max)) {
+      ref.current.setCustomValidity('Value greater than max');
+    } else {
+      ref.current.setCustomValidity('');
```
bachstatter, line 52: I think there is an edgecase bug here. If I put in an invalid number `abc` before max is fetched the input will have `Invalid number` validity. Then as max is fetched we reset the validity, and now we're in a state where the form thinks the input is valid eventhough it has `abc`

v3/components/NumberInput/NumberInput.tsx
```diff
@@ -32,20 +32,27 @@ export function NumberInput({
       try {
         nextValue = wei(e.target.value || 0);
       } catch (_err) {
-        e.target.setCustomValidity('Invalid number');
+        // whatever
```
bachstatter, line 35: Hmm without this the user will be allowed to submit `abc`?

v3/components/NumberInput/NumberInput.tsx
bachstatter: But it stops the modal from showing up and says invalid number.

v3/components/NumberInput/NumberInput.tsx
```diff
@@ -23,29 +25,46 @@ export function NumberInput({
...
+  useEffect(() => {
+    if (!ref.current) {
+      return;
+    }
+    if (!NUMBER_REGEX.test(`${inputValue}`)) {
+      ref.current.setCustomValidity('Invalid number');
+      return;
+    }
+    if (value && value.eq(0)) {
+      ref.current.setCustomValidity('Value required');
```
bachstatter, line 58: nice that was a mental note in the back of my mind

bachstatter (approved):

PR #702, opened 2023-02-09
Up

noisekit:
- Upgrade deps
- Skip i18next because their types are broken in a bad way in the new version
- Fix `disabled` -> `isDisabled` for all Chakra buttons. No idea why/how it worked before as chakra butttons never had `disabled` prop` (or why TS was happy about it)
- Manually resolve a few vulnerable dependencies


bachstatter (commented):
So is it the chakra update that requires `isDisabled` instead of `disabled`. Many of the buttons in v2 is still using `disabled` and that works

bachstatter (approved):

PR #704, opened 2023-02-12
Qa/round 1

jmzwar:


v3/components/CollateralTypeSelector/CollateralTypeSelector.tsx
```diff
@@ -38,7 +44,13 @@ export const CollateralTypeSelectorUI: FC<{
             onClick={() => onChange(collateral.symbol)}
           >
             <Flex flexDirection="row">
-              <CollateralIcon symbol={collateral.symbol} width="24px" height="24px" />
+              <CollateralIcon
+                fill="#0B0B22"
+                color="#00D1FF"
```
bachstatter, line 49: Given that this is used for all colalteral types I think the fill and the color should be set on the SNX icon instead

v3/ui/src/components/accounts/Deposit/DepositForm.tsx
```diff
@@ -234,8 +239,6 @@ export const DepositForm = (props: { staticCollateral?: boolean }) => {
       poolId={params.poolId}
       accountId={params.accountId}
       navigate={navigate}
-      DepositModal={DepositModal}
```
bachstatter, line 237: This dependency injection pattern is used so that the DepositFormUi can be end to end tested without netowrk request. We haven't been very consistent with this though

jmzwar: Ah I was wondering why this was there.

jmzwar: There were a couple of places IIRC

bachstatter: Yea, its needed for the component testing (not end to end) as I wrote before. I think it's also might be required to get storybook to play nicely

jmzwar: There is no other way to inject the dependency? It is a bit verbose having it in a few places.

bachstatter: Yea I dont know of a nicer pattern when UI components using a Container component. Could be explored. The other way would be to mock things in tests, but I'm not sure how flexible storybook is with mocking.

v3/components/CollateralTypeSelector/CollateralTypeSelector.tsx
jmzwar: Yeah, at the moment it doesn't affect the colour of say the eth icon because the props aren't passed. I think I'll create a separate PR - I want to expand that function and have a variant prop here

bachstatter (approved):

PR #728, opened 2023-02-20
Delegate mode

bachstatter:
Adds Delegate Mode.
When you in delegate mode you will see the wallet you acting on behalf on's data.
I've disabled link to the wallet page, when in delegate mode you can only: mint, burn or claim.
The user can always navigate to mint, burn and the claim page. If permissions are missing, the button will be disabled.

This now also hides the "Go back to old app" button. Which means we can remove a lot of code, but given that this PR is already big I didn't do it in this PR

https://user-images.githubusercontent.com/5688912/220203788-191c2f8d-6a81-4c6c-bd62-c6e8e0ecce02.mov


v2/components/WalletModal/WalletModal.tsx
```diff
@@ -46,23 +35,35 @@ export const WalletModalUi: FC<{
   ensName: string | null;
   walletAddress: string | null;
   networkId: number | null;
-  balances?: BalanceObject[];
+  Balances: FC;
```
jmzwar, line 38: Lets try to look deeper into a way around this pattern. It makes things difficult to read.

v2/lib/useAuthorisedWallets/useAuthorisedWallets.tsx
```diff
@@ -0,0 +1,66 @@
...
+    async () => {
+      const res = await fetch(url, {
+        method: 'POST',
+        body: JSON.stringify({
+          query: AuthorizedWalletsDocument,
+          variables: { delegate: walletAddress },
+        }),
+      });
+      const json: { data: AuthorisedWalletsQueryResult; errors: Error[] } = await res.json();
+      if (json.errors) {
```
jmzwar, line 45: Try catch would be nicer :P

jmzwar (commented):
Generally looks great.

v2/components/WalletModal/AuthorisedWallets.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { setDelegateWallet } = useDelegateWallet();
+  const navigate = useNavigate();
+  return (
+    <Box my={2} px={4} py={3} bg="black" border="1px" borderColor="gray.800" borderRadius="base">
+      <AuthorisedWalletsUi
+        isLoading={isLoading}
+        authorisedWallets={data}
+        onWalletSelected={(wallet) => {
+          setDelegateWallet(wallet);
+          navigate('/');
```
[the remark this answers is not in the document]
bachstatter: Yep I like that. Didn't use to have `onWalletSelected` from the parent. Fixed now

v2/components/WalletModal/AuthorisedWallets.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export type AuthorisedWalletsProps = {
+  onWalletSelected: (wallet: DelegateWallet) => void;
+};
+export const AuthorisedWallets: FC<AuthorisedWalletsProps> = ({ onWalletSelected }) => {
+  const { data, isLoading } = useAuthorisedWallets();
+  const { setDelegateWallet } = useDelegateWallet();
+  const navigate = useNavigate();
+  return (
+    <Box my={2} px={4} py={3} bg="black" border="1px" borderColor="gray.800" borderRadius="base">
```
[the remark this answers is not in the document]
bachstatter: done

v2/components/WalletModal/Balances.tsx
```diff
@@ -0,0 +1,117 @@
...
+                {formatNumber(balance)}
+              </Text>
+              <Text fontSize="xs" color="gray.800" textAlign="right">
+                {formatNumberToUsd(usdBalance)}
+              </Text>
+            </Flex>
+          </Flex>
+        );
+      })}
+      {delegateMode ? null : (
```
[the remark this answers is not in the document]
bachstatter: It would be `!delegate`. And if I do `!delegate` it would return `false` 
image
So UI would display `false` in general I'm a fan of using explicit teneraries instead of `&&` to avoid cases like this

v2/components/WalletModal/Balances.tsx
```diff
@@ -0,0 +1,117 @@
...
+      ? {
+          currencyKey: 'SNX',
+          balance: debtData.collateral.toNumber(),
+          usdBalance: debtData.collateral.mul(exchangeRateData.SNX || 0).toNumber(),
+          icon: <SNXIcon width="34px" height="34px" />,
+          description: 'Synthetix Network Token',
+        }
+      : undefined;
+
+  const synthBalances = synthBalancesData?.balances.slice(0, 5).map((x) => {
```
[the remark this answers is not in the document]
bachstatter: Hmm, what would be the purpose? Avoid a a rerender when something trigger an update? If any of the hooks in this contianers triggers an update I think we always want to rerender. I guess if all the balances are the same we would save an update. But gives that it's a map over 5 items I'm not sure how much performance improvements we would get.

Turning an array with objects to a key that useMemo can compare also isn't very easy

v2/components/WalletModal/WalletModal.tsx
bachstatter: I'm a dependency injection fanboy so it's not hard for me to read, but I guess it takes some getting use too. Happy for you to do a spike if you want

v2/lib/useAuthorisedWallets/useAuthorisedWallets.tsx
bachstatter: haha yea maybe. No way for us to change the Graph's api though

v2/components/WalletModal/Balances.tsx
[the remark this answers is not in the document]
jmzwar: Ok - if we always want to re-render that is fine

v2/components/WalletModal/WalletModal.tsx
jmzwar: Just feels like prop drilling.

This is specifically for component testing right? Do we need to component test the entire UI page?

jmzwar (approved):

v2/components/WalletModal/WalletModal.tsx
bachstatter: For UI testing and storybook. hmm it's sorta stems from wanting to avoid prop drilling. 
We could fetch everything in the `WalletModal` container and pass all the data to `WalletBalancesUI` from `WalletModalUi`. But to avoid prop drilling we create a `WalletBalances` container that fetch all the data `WalletBalancesUi` need.

bachstatter: In this case it would be fairly easy to rewrite it, but if `WalletBalances` was do be used elsewhere that component needs to contain the fetch logic too

jmzwar: RIght - in practice though - we shouldn't be testing something as large as a UI component in storybook. The UI level is our data fetching level, storybook should really be components a few layers down (which we can just pass props into in the story).

I think we are just testing too big things

PR #734, opened 2023-02-22
Fix query cache keys + assorted cleanups and fixes

noisekit:
These are extracted from my unwagmi refactoring (as not entirely related)
- Account may not exist (in the future, when wagmi is out)
- Simplify getting WETH price
- Fix missing React element key
- Cleanup old unused multistep txn modal
- Remove unnecessary Initialiser: we have Skeletons now and this one makes whole page flikr on load
- Align values and order of all the query cache keys, add missing network

### DIFF
--- a/v3/lib/useEthBalance/useEthBalance.ts
+++ b/v3/lib/useEthBalance/useEthBalance.ts
@@ -19,7 +19,7 @@ export function useEthBalance(networkId?: number) {
         networkId && networkId !== network.id
           ? new InfuraProvider(networkId, process.env.NEXT_PUBLIC_INFURA_PROJECT_ID)
           : connectedProvider;
-      return BalanceSchema.parse(await provider.getBalance(account?.address));
+      return BalanceSchema.parse(await provider.getBalance(account.address));
     },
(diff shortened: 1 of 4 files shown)


v3/lib/useEthBalance/useEthBalance.ts
```diff
@@ -7,20 +7,20 @@ import { InfuraProvider } from '@ethersproject/providers';
...
+    queryKey: [network.name, { accountAddress: account?.address }, 'EthBalance'],
     queryFn: async () => {
-      if (!accountAddress) throw Error('Query should not be enabled');
+      if (!account?.address) throw Error('Query should not be enabled');
       const provider =
         networkId && networkId !== network.id
           ? new InfuraProvider(networkId, process.env.NEXT_PUBLIC_INFURA_PROJECT_ID)
           : connectedProvider;
-      return BalanceSchema.parse(await provider.getBalance(accountAddress));
+      return BalanceSchema.parse(await provider.getBalance(account?.address));
```
bachstatter, line 22:
```suggestion
      return BalanceSchema.parse(await provider.getBalance(account.address));
```

bachstatter (approved):

PR #764, opened 2023-03-16
Refactor action data

bachstatter:
- Remove unnecessary useState and useEffect
- Remove nice action labels for AllAccounts table


bachstatter:
The code that keeps track of all positions would require us to have all data for every account for every asset, to make sure that the position is tracked correctly.. 
We cant gurantee that since things will be paginated from the graph. 
So I think we'd have to keep, position modified on the `AllAccountsTable`. Or move some of that logic to the subgraph

jmzwar (approved):

PR #766, opened 2023-03-17
Feature/mark data

jmzwar:


v2/perps-v2/ui/src/hooks/usePositions.ts
```diff
@@ -173,7 +181,17 @@ async function fetchPositions(positionData: PositionData[], address: string) {
   await Promise.all(
     positionData.map(async ({ entryPrice, leverage, market, asset }) => {
       const positionData = await contract.positionDetailsForMarketKey(market, address);
-      data.push({ position: positionData, market, leverage, entryPrice, asset });
+      const { fundingParameters, marketSizeDetails } = await contract.marketDetailsForKey(market);
```
bachstatter, line 184: Could promise.all. But maybe we're sending off enough requests at once either way

bachstatter (approved):

PR #768, opened 2023-03-17
Do one api call for all positions data

bachstatter:
Instead of 2 * number of positions api call, do one multicall.

It does get a lot harder to read, but I think it's worth it?


jmzwar (approved):
It's actually not too bad as you read through it

PR #775, opened 2023-03-20
Candles for perps subgraph

fritzschoff:


jmzwar (approved):
Looks good - I presume the startBlock changes won't affect historical data?

fritzschoff:
> Looks good - I presume the startBlock changes won't affect historical data?

correct, since we listen from an earlier point of time

PR #778, opened 2023-03-21
better prices + remove long/short on close

jmzwar:
Price was incorrectly fetched.

Now we return market.assetPrice() which is the oracle price.

Other minor changes.


v2/perps-v2/ui/src/components/Positions/PositionsTable.tsx
```diff
@@ -69,30 +68,44 @@ export const PositionsTable = () => {
...
+                const marketPrice = (parseInt(lastPrice) / 1e18) * (1 + skewRatio);
+                const sizeAmount = parseInt(size) / 1e18;
+                // Need to take away fees
                 const netValue =
-                  Math.abs(parseInt(size) / 1e18) * (parseInt(lastPrice) / 1e18) +
+                  Math.abs(sizeAmount) * marketPrice +
                   parseInt(funding) / 1e18 +
-                  parseInt(pnl) / 1e18;
+                  parseInt(pnl) / 1e18 -
+                  parseInt(fees) / 1e18;
```
bachstatter, line 85: Would be nice to extract out this calcs to a func and write some tests

v2/perps-v2/ui/src/hooks/usePositions.ts
```diff
@@ -227,8 +247,11 @@ async function fetchPositions(
         position: positionDetails,
         skew: marketSizeDetails.marketSkew,
         skewScale: fundingParameters.skewScale,
+        lastPrice: priceDetails.price,
```
bachstatter, line 250: So this is the current oracle price, do you think we can rename to to oracle price or index price? 
last price is confusing due to the abi using that in a weird way. And it could mean last interaction price (maybe)

v2/perps-v2/ui/src/hooks/usePositions.ts
```diff
@@ -150,8 +101,68 @@ export const usePositions = (walletAddress?: string) => {
...
               skewScale: skewScale.toString(),
+              fees,
             });
+          }
+        );
+
+        setState({ loading: false, data: update });
+      })();
+
+      const id = setInterval(() => {
```
bachstatter, line 112: a react-query with `{interval}` would simplify this function quite a bit

bachstatter (approved):
I left a bunch of comments that I think would improve things.. But if we want it out quickly go for it hehe

v2/perps-v2/ui/src/components/Positions/PositionsTable.tsx
```diff
@@ -69,30 +68,44 @@ export const PositionsTable = () => {
                   liquidationPrice,
                   skew,
                   skewScale,
+                  fees,
                 },
                 index
               ) => {
+                const skewRatio = parseInt(skew) / parseInt(skewScale);
+                // The last price is the price supplied by oracle.
```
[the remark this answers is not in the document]
jmzwar: Yeah v good point

v2/perps-v2/ui/src/hooks/usePositions.ts
jmzwar: For the position - yeah I think you are right. Index price should be oracle price and marketPrice should be with premium

PR #787, opened 2023-03-23
Position fixes

bachstatter:
This PR refactors things and fixes the PNL calcalulation.
The main refactor is that we now convert numbers to `wei` as soon as their fetched. I aslo changed all the  `<Td>` components to take number rather than strings. I also moved some calculation out of the components, especailly the mark price calc since that is need for PNL and a few other things.

The main logic for postion is now:
Grab data from the subgraph, to get open positions. The data we get back here is right up untill the point of the users most recent position modification. I've named things to reflect this.

We also grab position data from the contract. The data here is from the last position modification till now.

So with these two combined we can get the correct picture of PNL, funding etc. Happy to chat through things if needed.

One last thing to match with Kwenta is to use offchain prices rather than on chain. I think it make sense to do that in a separate PR


v2/perps-v2/ui/src/components/Actions/AllActionsTable.tsx
```diff
@@ -49,30 +48,27 @@ export const AllActionsTable = () => {
             {data?.map((item, index) => {
               const { label, address, asset, price, fees, size, txHash, timestamp, leverage } =
                 item;
-              const isLong = !size.includes('-');
+              const isLong = size.gt(0);
```
jmzwar, line 51: hehe much cleaner

jmzwar (approved):

PR #791, opened 2023-03-26
move contract logic to resolver

jmzwar:
Replace react-query with a client side resolver.

### DIFF
--- a/v2/perps-v2/ui/src/hooks/usePositions.test.ts
+++ b/v2/perps-v2/ui/src/hooks/usePositions.test.ts
@@ -4,7 +4,7 @@ import {
   calculateNewPnl,
   calculatePnlPercentage,
   calculatePositionData,
-} from './usePositions';
+} from '../utils/calculations';
 
 describe('calculateMarkPrice', () => {
   it('should correctly calculate the mark price', () => {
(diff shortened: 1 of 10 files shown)


v2/perps-v2/ui/src/utils/calculations.ts
```diff
@@ -0,0 +1,77 @@
+import Wei, { wei } from '@synthetixio/wei';
```
bachstatter, line 1: NIce.. It looks like tests not running in CI.. the usePostion.test.ts will fail now..

v2/perps-v2/ui/src/hooks/usePositions.test.ts
```diff
@@ -4,7 +4,7 @@ import {
   calculateNewPnl,
   calculatePnlPercentage,
   calculatePositionData,
-} from './usePositions';
+} from '../utils/calculations';
```
bachstatter, line 7: Nice, sorry missed this before

bachstatter (approved):
The resovlers being typed as `any` is very sad.. 
I quickly tried `@graphql-codegen/typescript-resolvers`, no success yet.. 
I guess we can live with it.. But if we cant figure it out I'd be in favour of adding zod schemas

PR #792, opened 2023-03-27
Feature/pyth realtime

jmzwar:
Add pyth offchain prices.


v2/perps-v2/ui/src/components/Positions/PositionsLoading.tsx
```diff
@@ -50,6 +50,11 @@ export const PositionsLoading = () => {
           Lorem Ipsum
         </Skeleton>
       </Box>
+      <Box as={Td} border="none">
+        <Skeleton startColor="gray.700" endColor="gray.900" my={2}>
+          Lorem Ipsum
```
bachstatter, line 55: Lorem -> Loading?

v2/perps-v2/ui/src/components/Positions/PositionsTable.tsx
```diff
@@ -75,7 +75,7 @@ export const PositionsTable = () => {
                   index
                 ) => {
                   // Need to take away fees
-                  const netValue = size.abs().mul(marketPrice).add(pnl);
+                  const netValue = size.abs().mul(marketPrice).add(pnl).sub(fees);
```
bachstatter, line 78: This should be moved to a calculation. Bu also pnl is gross and already taken fees into account.. So I dont think we want this change

v2/perps-v2/ui/src/hooks/usePositions.ts
```diff
@@ -58,7 +58,7 @@ export const usePositions = (walletAddress?: string) => {
     pollInterval: 1000,
   });
 
-  const positionData = (data?.positionsFromContract as PositionData[]) || undefined;
+  const positionData = data?.positionsFromContract as PositionData[];
```
bachstatter, line 61: The type here should be ` | undefined`?

v2/perps-v2/ui/src/utils/constants.ts
```diff
@@ -61,3 +61,249 @@ export enum FuturesMarketKey {
...
+  supports: 'mainnet' | 'testnet' | 'both';
+  version: 1 | 2;
+  pythIds?: {
+    mainnet: string;
+    testnet: string;
+  };
+  disabled?: boolean;
+}
+
+export const MARKETS: Record<FuturesMarketKey, FuturesMarketConfig> = {
```
bachstatter, line 77: I hadn't seen this file.. Ideally we should be fetching this this stuff rather than hardcoding. As it is now we need to remove/add as market gets added/removed

bachstatter (commented):
I do think separating price fetching and using websocket would be cool.. But this works for now. Some comments worth addressing

v2/perps-v2/ui/src/components/Positions/PositionsTable.tsx
jmzwar: Ok will remove

v2/perps-v2/ui/src/components/Positions/PositionsLoading.tsx
jmzwar: This is dummy text that never renders. It's just used to size the Skeleton

bachstatter (approved):

PR #796, opened 2023-03-28
NetFunding Rate update

fritzschoff:


v2/perps-v2/perps-subgraph/src/futures.ts
```diff
@@ -416,7 +439,7 @@ export function handlePositionModified(event: PositionModifiedEvent): void {
       .times(BigInt.fromI32(-1));
     const newPositionPnl = newPositionPnlWithFeesPaid
       .plus(futuresPosition.feesPaidToSynthetix)
-      .minus(futuresPosition.netFunding);
+      .plus(futuresPosition.netFunding);
```
bachstatter, line 442: It didn't cause a problem here.. Since this case handles positionOpened and netfunding will always be 0 when opening. Still good to have it looking right though :)

PR #820, opened 2023-04-20
removing candle entities

fritzschoff:


v2/perps-v2/perps-subgraph/subgraph.base.candle.yaml
```diff
@@ -0,0 +1,150 @@
+specVersion: 0.0.5
```
bachstatter, line 1: Did you add this file on purpose?

bachstatter: Why is this file called candle?

fritzschoff: cause it holds the candle related code that we dont use atm

fritzschoff: haha no wait
Screenshot 2023-04-21 at 09 25 25
this is what we got now. Candle is just a stash file

fritzschoff: base is what we use for deployment. Candle just holds the candle related code. Cause as mentioned in the description in this PR, if you comment out the candle entities, it will still end up as an endpoint that you can query and of course we dont want that

PR #825, opened 2023-04-26
Network agnostic e2e tests

noisekit:
- Switch e2e tests to use `optimism-mainnet`
- Update e2e tests to not hardcode actual network and retrieve it from provider instead. This means that whatever we fork - e2e will use that network. And we would be able to test on all supported networks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,7 +297,7 @@ jobs:
 
       - run:
           name: Run anvil localhost:8545
-          command: anvil --fork-url https://goerli.infura.io/v3/$INFURA_KEY
+          command: anvil --fork-url https://optimism-mainnet.infura.io/v3/$INFURA_KEY
           background: true
 
       - run:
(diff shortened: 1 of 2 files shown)


v3/cypress/cypress/tasks/getSnx.js
```diff
@@ -1,14 +1,29 @@
...
+  }
+}
 
 export async function getSnx({ address, amount }) {
   const config = await getCollateralConfig('SNX');
   const provider = new ethers.providers.JsonRpcProvider('http://127.0.0.1:8545');
 
+  const owner = await getOwner();
   //  const coreProxy = new ethers.Contract(CoreProxy.address, CoreProxy.abi, provider);
   //  const owner = await coreProxy.owner();
```
bachstatter, line 28: Could this be enabled now? Dont remember why this is commented out

bachstatter (approved):

PR #826, opened 2023-04-26
Allow switching networks when wallet not yet connected

noisekit:
- After page refresh selection is preserved
- If walled connected and network switched inside wallet - that is preserved too (and will be restored if wallet is disconnected)

image


bachstatter (approved):

bachstatter:
Needs a deps

PR #856, opened 2023-05-09
Add perps-watcher IPFS deployment

noisekit:


bachstatter (approved):
Should we move ipfs-deploy to Synthetix npm. Can put it in packages?

PR #886, opened 2023-05-24
added pendingnode id

fritzschoff:


bachstatter (approved):
Agree with Nikitas comments, no blockers though

PR #894, opened 2023-05-29
Upgrade wallet connect

bachstatter:
I've added `NEXT_PUBLIC_WC_PROJECT_ID` to fleek and vercel.

These are the whitelisted domains:
image

To configure those you log in here: https://cloud.walletconnect.com/
Creds in 1pass


bachstatter:
Still running into some problems, so keeping this in draft for now

PR #902, opened 2023-05-30
V3 move

noisekit:
- All the v3 and actively worked on code is now moved to https://github.com/Synthetixio/v3ui
- All moved filed are removed here to avoid any duplication of work
- This monorepo is going to be used only for bugfixes/support


bachstatter (approved):
I didn't spot anything weird

PR #913, opened 2023-06-08
only show gas estimate on mainnet

jmzwar:


bachstatter (approved):
I think mainnet would have similar discrepancies.. But I dont want to investigate so this looks good 😅

PR #918, opened 2023-06-14
add steth

jmzwar:


bachstatter (approved):
Need an icon too

jmzwar:
> Need an icon too

Yup jade is on it

PR #923, opened 2023-06-15
Do not require any chains

noisekit:
WV2 try 2


bachstatter (approved):
No sure if we wanted this or not...

PR #953, opened 2023-09-05
Feature/search ens

Steve-Rog:


v2/perps-v2/ui/src/components/AddressInput/AddressInput.tsx
```diff
@@ -12,14 +15,23 @@ export const AddressInput = () => {
...
+      (async () => {
+        const addressFromEns: string | null = await L1DefaultProvider.resolveName(
+          getValues('address')
+        );
+        if (addressFromEns) {
+          navigate(`/${addressFromEns}`);
+        } else {
+          navigate(`/${getValues('address')}`);
+        }
+      })();
```
bachstatter, line 27: Can you use ethers.utils.isAddress and only do the ens lookup is the value isn't an address?

bachstatter: Also I think it's fine to just make onSubmit an async fn rather than having an IIFE inside

v2/perps-v2/ui/src/hooks/useENSName.ts
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react';
+import { useGlobalProvidersWithFallback } from '@snx-v2/useGlobalProvidersWithFallback';
+
+const fetchENSNameForAddress = async (
+  address: string | null | undefined,
+  provider: any
```
bachstatter, line 6: `import {providers} from 'ethers'`
`any` -> `providers.JsonRpcProvider`

bachstatter (commented):
Hey! Thanks for the PR. Left some minor comments

PR #955, opened 2023-09-08
Watcher auto updates

bachstatter:
This PR fetches pyth ids from `offchain-feeds.json` instead of hardcoding them
Also needed a marketSummaries hook.

With all the new markets I was also running into 413 from the subgraph. This means our reqeuest payload was too large, so I removed fields we we were not using


bachstatter:
> How about limiting graph response to like 100 rows or smth?

It was actually not the response that was the problem, they do add pagination and have a max page size.. 
This error we ran into was the *request* query being too big, eg not the response.

jmzwar (changes requested):
This is good @bachstatter however we also need the contract addresses for each market to be able to fetch them for the market page.

v2/perps-v2/ui/src/hooks/useMarketSummaries.ts
```diff
@@ -0,0 +1,18 @@
...
+import { perpsMarketDataContract } from './usePositions';
+
+export const useMarketSummaries = () => {
+  const [data, setData] = useState<{ asset: string; address: string }[]>([]);
+  useEffect(() => {
+    perpsMarketDataContract
+      .allProxiedMarketSummaries()
+      .then((res) =>
+        setData(
+          res.map((x) => ({ asset: parseBytes32String(x.asset), address: x.market.toLowerCase() }))
```
bachstatter, line 12: @jmzwar  That's why I added this hook

jmzwar: Hmm ok - I didn't see the new markets on the markets page. Checking now

jmzwar: Ok I think we just don't have enough historical data yet. Gtm

jmzwar (approved):

PR #958, opened 2023-09-12
Feature/kwenta poly

Steve-Rog:
Merge of a user's actions, signaling which of them are derived from the KWENTA protocol and which from POLYNOMIAL.

Screenshot 2023-09-18 alle 09 46 38


v2/perps-v2/ui/src/hooks/useMergedActions.ts
```diff
@@ -0,0 +1,50 @@
...
+  } = useActions(kwentaAccount?.account || '');
+  const {
+    data: dataPolynomial,
+    loading: loadingPolynomial,
+    error: errorPolynomial,
+  } = useActions(polynomialAccount?.account || '');
+
+  const isDataLoading = loading || loadingKwenta || loadingPolynomial;
+
+  let mergedArray: { [key: string]: any }[] = [];
```
bachstatter, line 23: You should now the type here? Prefer if we dont use `any` here... Since I think this will make the return data to be `any` as well?

bachstatter:
Does this mean we can remove the Kwenta/ Poly links too?

v2/perps-v2/ui/src/hooks/useMergedActions.ts
Steve-Rog: committed changes

jmzwar:
@Steve-Rog closing this

PR #974, opened 2023-10-03
Open positions Table w/ Filtering

keybig:
grants council task: "open position leaderboard (real time updates) for largest/smallest: pnl/size, asset filter"

Adds a `Positions` route and page, that renders a table of the top open positions, with options for sorting and filtering the Open Positions by 'Size', 'Unrealized PNL', and 'Realized PNL'. Also handles pagination.

image

Updated the `usePositions` hook to be able to fetch all open positions, and filter/sort/paginate those positions based on url parameters. The hook now can handle fetching all open positions in addition to fetching the open positions for a single address.

Created a Shared Component for Dropdowns, that is being used for the `OrderBy` and `OrderDirection` components. 

Updated the `MarketSelect` component to accept a `route` prop, so that it can be used on different pages and accurately update the URL params. Also updated the handling on URL params to make them dynamic. 

Created OpenPositions components, based off existing Positions components, to allow users to filter open positions.

### DIFF
--- a/v2/perps-v2/ui/src/components/Header.tsx
+++ b/v2/perps-v2/ui/src/components/Header.tsx
@@ -96,13 +96,20 @@ export const Header: FC = () => {
                     </Text>
                   </RouterLink>
                 </MenuItem>
+                <MenuItem>
+                  <RouterLink to="/positions">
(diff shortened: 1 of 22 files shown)


jmzwar (approved):
Great work guys

PR #999, opened 2023-10-25
websocket support

jmzwar:


v2/perps-v2/ui/src/utils/pyth.tsx
```diff
@@ -51,3 +54,40 @@ export const getMarketsPythConfig = () => {
...
 };
+
+export interface PythPrice {
+  price: string;
+  conf: string;
+  expo: number;
+  publishTime: number;
+}
+
+export const prices: { [key: string]: PythPrice } = {};
```
bachstatter, line 65: Usually I get scared when I see state in an exported const, not sure if it's an actually problem, @noisekit  might know?

bachstatter: This also means that we dont get any rerender when it updates. Is that what we want since the other hooks runs on an interval?

v2/perps-v2/ui/src/utils/pyth.tsx
```diff
@@ -51,3 +54,40 @@ export const getMarketsPythConfig = () => {
...
+        if (priceData) {
+          prices[id] = priceData;
+        }
+      });
+    })();
+
+    return () => {
+      // Clean up WebSocket connection when the component unmounts
+      (async () => {
+        const pythConfigByMarketKey = await getMarketsPythConfig();
```
bachstatter, line 85: We can use `Object.keys(prices)` to avoid this await

v2/perps-v2/ui/src/utils/pyth.tsx
jmzwar: Correct

bachstatter: Yea still feel like it's could cause problems, but if it's working fine I guess we can ship it. Did you see my other comment too?

v2/perps-v2/ui/src/utils/pyth.tsx
jmzwar: done

v2/perps-v2/ui/src/utils/pyth.tsx
jmzwar: I actually prefer it this way as the socket can't affect the render cycle (only the refetch does).

jmzwar: What did you have in mind by problems?

bachstatter: I guess this would be hard to test and I would be more use to seeing it controlled with memorisation, only rerender if values change. Feels pretty opinionated so shouldn't be a blocker

bachstatter (commented):
No need for this async function

jmzwar:
> No need for this async function

which one?

bachstatter (commented):
Hmm yea,I find it a little annoying if I paste in a url and the back button takes me to the website I was previously on. Haha happens all the time on DeBank.

bachstatter (approved):

bachstatter (commented):
This one async

PR #1010, opened 2023-11-06
incremental

jmzwar:


v2/lib/formatters/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../../tsconfig.base.module.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "rootDir": "./"
+    "rootDir": "./",
+    "incremental": true
```
noisekit, line 6: tbh i'd really suggest not to do incremental. these libs are pretty small, and incremental builds do cache things and tend to have issues with cache (when publishing especially)

noisekit (approved):

PR #1042, opened 2024-01-15
0xwalde add debt migration UI

jmzwar:

PR #1045, opened 2024-01-16
0xwalde add debt migration UI

noisekit:
Reopen to fix problems of #1042


v2/components/WalletLayout/WalletLayout.tsx
```diff
@@ -29,13 +29,20 @@ const WalletLayoutUi = ({
   children,
 }: PropsWithChildren<{ networkId: number | null }>) => {
   const { t } = useTranslation();
-  const isMainnet = networkId === NetworkIdByName.mainnet;
+  const isMainnet = networkId === NetworkIdByName.mainnet || networkId === NetworkIdByName.goerli;
```
noisekit, line 32: Goerli is not a mainnet!

v2/ui/sections/migrate/components/TabContent.tsx
```diff
@@ -39,10 +42,12 @@ type TabContentProps = {
   isVestNeeded: boolean;
   resetTransaction: () => void;
   optimismLayerOneFee: Wei | null;
+  type: 'escrow' | 'debt';
```
noisekit, line 45: This should gave been copied and split into two separate components MigrateEscrow and MigrateDebt. And existing migrate escrow should have stayed untouched. Have we tested escrow migration after this change?

v2/ui/sections/migrate/components/TxSent/ActionCompleted.tsx
```diff
@@ -13,6 +13,7 @@ type ActionCompletedProps = {
   resetTransaction: () => void;
   amount: string;
   action: 'migrate';
+  type: 'escrow' | 'debt';
```
noisekit, line 16: Same here. Split into 2 independent components

v2/ui/sections/migrate/components/TabContent.tsx
```diff
@@ -70,10 +74,10 @@ const TabContent: FC<TabContentProps> = ({
           size="lg"
           disabled={false}
         >
-          {t('migrate-escrow.actions.migrate.action.go-to-escrow-page')}
+          {t(`migrate-${type}.actions.migrate.action.go-to-escrow-page`)}
```
noisekit, line 77: Template strings must not be templated, it is impossible to find them with code search.

noisekit (commented):
This needs refactoring it cannot be merged like this. No existing code must be changed, only new one added, this way we avoid regressions.

v2/components/WalletLayout/WalletLayout.tsx
bachstatter: Probably just call the const L1 instead

0xwalde: Ah yep, I meant to call this L1. Will fix

v2/ui/sections/migrate/components/TabContent.tsx
0xwalde: Got it, I wasn't sure what the preference was here since many of the components are 90% the same between escrow and debt migration. I did test the escrow migration page as well after the change, but I can break it back out into completely separate sets of components if total isolation is preferred 👍

v2/ui/sections/migrate/components/TabContent.tsx
0xwalde: Okay, will change-- the only reason I did it was that I saw it done elsewhere in the codebase. But once this is split back into separate debt/escrow versions it won't be necessary anyway :+1:

bachstatter (approved):
Looks good, didn't do any testing though

PR #1056, opened 2024-02-01
fix close review withdraw modal

duyentb95:


v2/ui/components/NativeBridge/ReviewWithdrawModal.tsx
```diff
@@ -118,10 +119,12 @@ export const ReviewWithdrawModal: FC<{
               status: 'success',
             });
           }
+
+          await sleep(10 * 1000); // Delay 10s before closing modal
```
noisekit, line 123: timed close of modals is always a very bad idea, closing should happen on action or state change.

duyentb95: Removed it -> Allows users to manually close the modal after show successful status

jmzwar (approved):

v2/ui/components/NativeBridge/ReviewWithdrawModal.tsx
noisekit: Thanks

